PR #29, opened 2016-04-18
use git submodule instead of git clone when install

chungth:
I think we should to use `git submodule` here, because if I use `git clone `, and add `docker` folder to my project git repo, when my teammate pulls from  remote, he will faced the `No submodule mapping found in .gitmodules for path 'docker'` error.


Mahmoudz:
@chungth thanks

Issue #10, opened 2016-01-29
Support Postgres


unknownArtist:
can we use Postgres right now with laradock ?

Mahmoudz:
Not yet, but I would be more than happy if you would contribute with that.

PR #34, opened 2016-05-07
Add support for postgres

benmag:
Adds postgres support #10


Mahmoudz:
@benmag I was trying postgres and I got this error: 

```shell
  [PDOException]
  SQLSTATE[HY000] [2002] Connection refused
```

How did you managed to fix that on your machine?

I think we are missing something like `pdo-pgsql` from the php dockerfile!
Can you advise please, thanks in advanced.

benmag:
You are correct, we are missing the `pdo_pgsql` extension.

I'll send another PR to fix it.

PR #44, opened 2016-05-12
Memcached support

mattythebatty:
Add memcached support to laradock.

Installing memcached for PHP7 is a bit tricky. As far as I can see the other alternatives are just as unwieldy so you might want to comment it out and let users decide.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,6 +34,7 @@ services:
             - /var/lib/postgres
             - /var/lib/mariadb
             - /var/lib/redis
+            - /var/lib/memcached
 
 ### MySQL Container #########################################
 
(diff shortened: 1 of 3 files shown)


Mahmoudz:
@mattythebatty Hi again, hope you are doing well.
Before you supported Memcached (295a0974dafb14c35bc7836b3e020bd84d726ae7) I was letting users to switch the PHP version my editing the PHP `Dockerfile` Base Image. However after you add the support you've included some libraries that just works for PHP 7 which let me think in a better solution that keeps running for all the PHP versions. So I came up with multiple `Dockerfiles` each for a PHP version to prevent any similar issues with other version specific libraries in the future. 

For that I created a new release 1.4.0 (405067d4afaf8d0c62ced33d3716ba98cd101ed0) with 3 differenet `Dockerfiles` however I would like to ask you, when you have some free time to add the `memcached` libraries for the PHP `Dockerfiles` of the version 5.6 and 5.5 which can be found here https://github.com/LaraDock/laradock/tree/master/php

Thanks again

mattythebatty:
Hi @Mahmoudz , I've created a PR for the PHP5 docker files. It's a lot simpler than the PHP 7 version.

Thanks for creating this project.

Mahmoudz:
@mattythebatty thanks for your contributions 👍

PR #52, opened 2016-05-18
Install node, npm, gulp and bower to the workspace container

mattythebatty:
Not sure if you want it to work this way but there's some extra tools here. Tasks like gulp seem to work well but you don't get the little notification.


Mahmoudz:
@mattythebatty thank you bro :)

Issue #53, opened 2016-05-18
Problem on apt-get update


cabrerabywaters:
Hello there Mahmoud!

I'm having an issue while trying to set up Laradock.

When I do a 
`docker-compose up -d  nginx mysql redis workspace`

I'm getting an error on the apt-get update command on line 15 of the Workspace Dockerfile.

> W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/trusty-updates/main/source/Sources  Hash Sum mismatch

I tried running 
`rm /var/lib/apt/lists/* -vf`
in the Dockerfile before the update, but I'm still getting the same error.

![screen shot 2016-05-18 at 8 42 02 pm](https://cloud.githubusercontent.com/assets/3051507/15378339/0abd825e-1d39-11e6-968f-b31328e7bb0c.png)

Any ideas?

Mahmoudz:
Hi, I rebuilt the Workspace container twice on my machine and one once a Vagrant machine, and wasn't able to reproduce that error.
What I advice is to this:
1. Pull the latest version: `git pull origin master` now it's `2.2.0`
2 . Remove the workspace images and any `<none>` images with `docker rmi {container-id}`
3 . Rebuild the workspace `docker-compose build --no-cache workspace`
4 . Try again `docker-compose up -d nginx mysql redis workspace`

cabrerabywaters:
Thank you Mahmoud, it worked!docker-machine l.

I think that maybe doing the 

`git submodule add https://github.com/LaraDock/laradock.git`

as written in the Readme, did not pull the latest version of the repo

Mahmoudz:
@cabrerabywaters it does actually but before answering your question I released new version was working on.
Best,

matriphe:
Found the same problem. Running on OSX 10.11.4, using Docker Toolbox (Kitematic) version 0.10.2 that runs docker version 1.11.1 and docker-machine version 0.7.0.

Have tried to run the suggested step, but it's failed on `docker-compose build --no-cache workspace`

The error:
```
W: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/trusty/universe/binary-amd64/Packages  Hash Sum mismatch

E: Some index files failed to download. They have been ignored, or old ones used instead.
ERROR: Service 'workspace' failed to build: The command '/bin/sh -c apt-get update && apt-get install -y     software-properties-common' returned a non-zero code: 100
```

Mahmoudz:
[...]

PR #59, opened 2016-05-22
Replacing phusion/baseimage to sunfoxcz/baseimage

matriphe:
sunfoxcz/baseimage is based on Ubuntu Xenial 16.4 since the official phusion/baseimage has not support it yet. It's also referencing issue #53

### DIFF
--- a/beanstalkd/Dockerfile
+++ b/beanstalkd/Dockerfile
@@ -1,4 +1,4 @@
-FROM phusion/baseimage:latest
+FROM sunfoxcz/baseimage:latest
 
 MAINTAINER Mahmoud Zalt <mahmoud@zalt.me>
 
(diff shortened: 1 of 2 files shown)


Mahmoudz:
@matriphe thanks for your contribution. I'm glad this works for you. 

I don't think it's recommended to use `apt-get upgrade` since it takes a lot of time!
And I prefer to keep using `phusion/baseimage` as it's widely supported. Will address that issue in a different way without changing the base image.

Thanks again and keep considering sending PR's.
Regards,

matriphe:
Yes. I agree to wait the official support of `phusion/baseimage` for Ubuntu Xenial but for now it worked for me. thanks! 

FYI, the `apt-get upgrade` on the freshly Ubuntu Xenial (docker) only upgrade 5 packages, `dpkg` `libsystemd0` `libudev1` `systemd` `systemd-sysv` and it took lesser time rather than `apt-update`-ing.. :D

PR #60, opened 2016-05-22
Update README.md

AminMkh:
No need to run 2 separate commands for chmod..

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ DB_HOST=xxx.xxx.xxx.xxx
 **Debugging**: in case you faced an error here, run this command from the Laravel root directory:
 
 ```bash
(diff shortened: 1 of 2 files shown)


AminMkh:
Also, regarding, docker-compose.yml
I use the machine name as the DB_HOST, this way it's dynamic and not hardcoded.

```
DB_HOST=mysql
DB_DATABASE=homestead
DB_USERNAME=homestead
DB_PASSWORD=secret
```

Mahmoudz:
@AminMkh I see this PR contains the PR #61.

This command is so nice `sudo chmod -R 777 storage bootstrap/cache` 

Let's split both requests it will be easier for me to merge them. 
I'm waiting your PR with only the above update.

Thanks.

PR #61, opened 2016-05-22
 Updated docker-compose.yml

AminMkh:
I use the machine name as the DB_HOST, this way it's dynamic and not hardcoded.

```
DB_HOST=mysql
DB_DATABASE=homestead
DB_USERNAME=homestead
DB_PASSWORD=secret
```

I don't need to setup an IP, please test this out, it could be docker toolbox related, not sure I use linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ DB_HOST=xxx.xxx.xxx.xxx
 **Debugging**: in case you faced an error here, run this command from the Laravel root directory:
 
 ```bash
-sudo chmod -R 777 storage && sudo chmod -R 777 bootstrap/cache
+sudo chmod -R 777 storage bootstrap/cache
 ```
 
 <br>


Mahmoudz:
Hi @AminMkh first thank you for this contribution. using mysql as host instead of the docker IP is a very cool, I will update the readme and mention that as an option.

Regarding this PR you have switched the dependencies between the containers, this actually work but in that case when you run an NGINX container it will also run a MySQL and that's a problem when you are using other DB engine (example: if you want to use Postgres with NGINX it will also run MySQL! since there's a link between PHP-FPM "dependency on NGNIX" and MySQL) .

However I was wondering as well why to have this relation from the beginning!! so I tested MySQL, NGINX and PHP-FPM and turns out it's not necessary to have a relation between them at all :D which is cool. 

If you could please send another pull request removing the relations "links" between PHP-FPM and all the DB engines (mysql, postgres, mariadb and neo4j) would be great. Not sure if Redis and Memchached also require that link if you can test them also consider removing those relations.

Looking forward for the updates..

PR #64, opened 2016-05-23
Update docker-compose.yml

AminMkh:
It seems that since docker-compose yaml version 2, it doesn’t require a link to other containers in order to connect to them, it will resolve the image name into an IP by default. Though not clearly mentioned in docker-compose documentation !


Mahmoudz:
@AminMkh Thanks that looks more clean :)

PR #71, opened 2016-05-24
Prepare Workspace for multiple php versions

Gummibeer:


Mahmoudz:
@Gummibeer since the workspace container is not PHP centric, and it contains many tools where each tool has different version, I prefer not to do the suggested change (at least for now). For now if users want to change the PHP version or any other tool version they have to manually modify the dockerfile of the workspace.
Thanks.

Issue #104, opened 2016-06-20
Add Prestissimo composer plugin in workspace image


luciano-jr:
This plugin (https://github.com/hirak/prestissimo) help us with composer to have a parallel install approach.

It's very useful, the results makes really difference in the workflow.

Mahmoudz:
@luciano-jr Yes I'm familiar with it. Do you think we should have it by default installed on the Workspace container!! or let interested developers installed it on their own images whenever needed!

luciano-jr:
@Mahmoudz I don't think we will have problem to add by default in the container. The key here is that most developers don't know about Prestissimo, maybe we could just recommend in the documentation or something like this.

PR #114, opened 2016-06-22
Added support to conditionally install Prestissimo

Kauhat:
This is a bit of an odd one, so let me know what you think.

I've added an `INSTALL_PRESTISSIMO` argument to the workspace dockerfile which will conditionally run the relevant composer install if set to `true`. I like that idea it would only be installed if wanted, although in order to do this I'm running a bash `if` in the dockerfile itself - which could be considered a bit of a hack. Apart from this, it seems to be working well. It runs right at the end, so the cache for the rest of the container should never be invalidated.

One thing to note at the moment is that if no value is supplied you will receive a warning message when building. Is there any way to supply a default value in the `docker-compose.yml` file?

To try this out, create a `.env` file in the main directory and paste the below:

```bash
INSTALL_PRESTISSIMO=true
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /logs
+.env
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -153,7 +153,10 @@ services:
 ### Workspace Utilities Container ###########################
 
     workspace:
-        build: ./workspace
+        build:
+            context: ./workspace
+            args:
+                INSTALL_PRESTISSIMO: ${INSTALL_PRESTISSIMO}
         volumes_from:
             - application
         tty: true
--- a/workspace/Dockerfile
+++ b/workspace/Dockerfile
@@ -88,3 +88,9 @@ RUN . ~/.bashrc
 RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 WORKDIR /var/www/laravel
+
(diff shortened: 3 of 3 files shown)


Kauhat:
#114

Kauhat:
#104

Mahmoudz:
@Kauhat That looks fantastic, thank you. I think we should do this for many other tools in the future, allowing users to choose what they need..

However, is there a way to set the default value (`false`) in the `docker-compose` file? If yes please submit a PR with it otherwise let's ship a default `.env` (or `.env.default`) file with the repo.

Also can you please add a section in the README about this. Title could be: "Enable `Prestissimo` in the `workspace` Container". In the section just add 1 line about what is `Prestissimo` then how to turn it ON and OFF.

Thanks again.

luciano-jr:
That is awesome! Thank you @Kauhat 👍

Kauhat:
@Mahmoudz I could never figure this on our I'm afraid - setting a default value using the below prevents the warning, but the flag in the `.env` file doesn't seem to work.

```
    workspace:
        build:
            args:
                - INSTALL_PRESTISSIMO="false"
    ...
```

If I figure it out, I'll let you know, as it may be useful for other things later on.

Mahmoudz:
@Kauhat I'll leave it like this till someone figures out how to do it.. Thanks

PR #126, opened 2016-06-30
Added ruby container

zhikiri:


zhikiri:
Added supporting of the ruby container

PR #171, opened 2016-07-22
Created Alpine PHP FPM and Workspace containers

Kauhat:
Here's some Alpine Linux PHP and Workspace containers. These are built differently to the standard containers, as they use the system package manager to install the extensions (faster build, no compilation issues). They're also a lot smaller! I've added to the [change the php fpm version](https://github.com/Kauhat/laradock/blob/7f41ffb863e0f10d4e8c2962d07baf37388aaa27/README.md#change-the-php-fpm-version) section in the README to illustrate how these are enabled.

I've been using these for about a week now and haven't run into any noticeable issues but your milage may vary!

I've also removed the `INSTALL_PRESTISSIMO` argument and added a `./workspace/composer.json` file instead - this replaces the original functionality while being more extendable and easier to update.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*]
+charset = utf-8
+
+[{Dockerfile,docker-compose.yml}]
+indent_style = space
+indent_size = 4
--- a/README.md
+++ b/README.md
@@ -874,10 +874,13 @@ Additionally, you can contact Mahmoud Zalt (the creator of this project) via a d
 
 **Main Contributors:**
 
-- [Jack Fletcher](https://github.com/Kauhat)
-- [Bo-Yi Wu](https://github.com/appleboy)
-- [Amin Mkh](https://github.com/AminMkh)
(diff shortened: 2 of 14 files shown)


Mahmoudz:
@Kauhat first of all thank you for all your contributions.

You have done some impressive work especially the global composer thingy.

I've always wanted to make the workspace and php containers work faster and I had some plans for them but never found time!!

My plan was to create an images as bases Image for each of those containers and built from them. This will definitely speeds up the building time.

However, your alternative will also work and maybe doing both methods will speeds up things even more. 

My only concern for merging this PR is that we'll have 2 workspaces and almost 6 php containers. Which will be very hard to maintain and most importantly people will get lost.. 

Since your main goals are very convenient (faster build and less compilation issues) let me try to couple things in the next few days and compare the results to yours..

I should end up using one of those two base images: Alpine or the original PHP.. and get back to you soon.

Best,

Mahmoudz:
I hope LaraDock 4 will speed up the build and eliminate compilation issues.

Kauhat:
I'm really liking the changes you made in 3.0.0 and up @Mahmoudz!

PR #174, opened 2016-07-24
Add hhvm

orette:
Added a container running HHVM FastCGI. 
To use the container change the variable PHP_PROCESS_MANAGER to equal hhvm and rebuild nginx. #


Mahmoudz:
@orette this is amazing people have been waiting for this feature for a long time. Thank you.

PR #184, opened 2016-07-26
Nginx alpine image

### LINKED ISSUE #119: default conf
When running laradoc's nginx, the default conf is taken instead of laradock.conf
this could be solved either by deleting `/etc/nginx/conf.d/default.conf` in Dockerfile or by incuding only upstream.conf in nginx.conf:24 instead of *.conf

computerfr33k:
* Removed default site config to enable browsing from localhost address.
* Closes #119 

Let me know if you would rather have it as an alpine image instead of being the default, or any other thoughts you might have related to this.

### DIFF
--- a/nginx/Dockerfile
+++ b/nginx/Dockerfile
@@ -5,9 +5,13 @@ MAINTAINER Mahmoud Zalt <mahmoud@zalt.me>
 ADD nginx.conf /etc/nginx/
 ADD laravel.conf /etc/nginx/conf.d/laravel.conf
 
-ARG PHP_UPSTREAM
+ARG PHP_UPSTREAM=php-fpm
 
-RUN rm /etc/nginx/conf.d/default.conf \
+RUN apk update \
+    && apk upgrade \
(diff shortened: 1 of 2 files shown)


Mahmoudz:
@computerfr33k No problem switching to NGINX Alpine. Can you confirm this is working just fine (please test it with a fresh Laravel installation and check if you can see the welcome page)? Also could you please reference what Issues this fixes.. Thanks in advanced.

Kauhat:
Working great on my end!

```
/ # cat /etc/*-release
3.3.3
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.3.3
PRETTY_NAME="Alpine Linux v3.3"
HOME_URL="http://alpinelinux.org"
BUG_REPORT_URL="http://bugs.alpinelinux.org"
```

computerfr33k:
I can confirm that it works with a fresh laravel project & installation.
I have also put in a fix because it wasn't actually logging to the log files when the access_log was set to "on".

PR #191, opened 2016-07-28
parameter data in the path

Selsynn:


Kauhat:
Some previous discussion on this: 68576f9022df23bf9f28cf9f17855aa207bd8e70

Mahmoudz:
I've tried this now again and still cannot get it to work with `./data` on the Docker VM.
What do you guys suggest for supporting both Docker VM and Docker Beta?! 
Cause I think at least 60% of LaraDock users are still using the Docker VM!!

Kauhat:
I'm using `docker-machine` at home and the new beta at work and I've not had any problems with the above on either installation. Has it been the same VM causing problems?

Additionally, it might be worthwhile to add the `:z` label to each of the volumes, as these are shared between containers.

[Some more info](https://docs.docker.com/engine/tutorials/dockervolumes/#/volume-labels)

Selsynn:
I'm on docker vm and i cannot use /var/lib because we are on mac. But with data on the same project, there is no problem.
I think there are something in your configuration of docker

By the way i ignore how to cancel the pull request, i wanted to pull request on my repo, because i understand the problem.

PR #221, opened 2016-08-11
Added PHP Memcached extension installation.

tianissimo:
PHP memcached extension is required but it has been missing.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,6 +26,7 @@ services:
             args:
                 - INSTALL_MONGO=false
                 - INSTALL_XDEBUG=false
+                - INSTALL_MEMCACHED=false
             dockerfile: Dockerfile-70
         volumes_from:
             - volumes_source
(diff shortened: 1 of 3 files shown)


Mahmoudz:
Hi @tianissimo that looks great. 

We have `Memcached` installed on the [base Image](https://github.com/LaraDock/php-fpm/blob/master/Dockerfile-70) of the PHP 7 image only "and it looks very way".

I would love to remove the old Memcached installation code from the base image and merge your PR instead. Can you just test it (with `PHP 7`) and confirm it's working correctly?

Thanks, in advanced.

tianissimo:
I tried it and found that it's not working with PHP 7. pecl/memcached requires PHP version >= 5.2.0 and <= 6.0.0.
Maybe, we have to use existing method only for PHP 7.

Mahmoudz:
@tianissimo true, could you please send another PR with the changes only on PHP 5.6? 

After I merge you PR, I will get the memcached code from the PHP 7 base Image and make it optional on the PHP 7 (FPM Container) here, so the users can specify whether to install it or not on both PHP versions. 

Thanks.

PR #231, opened 2016-08-15
Proof of concept - LaraDock hosting multiple sites

LarryEitel:
Proof of concept - LaraDock hosting multiple sites

PR #238, opened 2016-08-16
Multi site framework agnostic

tjb328:
This will allow laradock to become framework agnostic building on PR #231 . This adds a folder for example conf files which contains a drupal 8 example. This will also install opcache in the PHP-FPM container and Drush in the workspace container.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,6 +11,7 @@ services:
                 - INSTALL_MONGO=false
                 - INSTALL_XDEBUG=false
                 - INSTALL_NODE=false
+                - INSTALL_DRUSH=false
                 - COMPOSER_GLOBAL_INSTALL=false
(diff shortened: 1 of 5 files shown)


LarryEitel:
@tjb328 are you available on [gitter](https://gitter.im/LaraDock/laradock)?

LarryEitel:
What do you think about:

```
    workspace:
        build:
            context: ./workspace
            args:
                - INSTALL_MONGO=false
                - INSTALL_XDEBUG=false
                - INSTALL_NODE=false
                - INSTALL_DRUPAL=false
```

That way all that Drush stuff won't be dumped on Laravel Devs. 

Thoughts?

tjb328:
Good idea. Let me get another commit put together. I'll also include some changes for opcache.

PR #257, opened 2016-08-26
Create README-zh.md

zhqagp:
I am a Chinese engineer, just translate your README, I want to let more Chinese people to use this product easily?


Mahmoudz:
@zhqagp that's fantastic. 谢谢

PR #269, opened 2016-08-30
expose-port-22-and-add-insecure_id_rsa_ppk

LarryEitel:
corrected placement of ports


oprudkyi:
This commit adds second 
```
        ports:
           - "22:22"
```

into docker-compose.yml and make all package unusable
ERROR: for php-fpm  driver failed programming external connectivity on endpoint laradock_php-fpm_1
Error starting userland proxy: listen tcp 0.0.0.0:22: bind: address already in use

LarryEitel:
@oprudkyi Exposing this port is important for remote debugging. If it conflicts, just map it to an unused port, i.e., "2200:22". Plus you may want to connect via a terminal other than docker-bashing into it. :)

oprudkyi:
I just pointed out that https://github.com/LaraDock/laradock/blob/master/docker-compose.yml have two services which uses port 22 , so without editing of docker-compose.yml by every new user the docker will complain

LarryEitel:
@oprudkyi You are right. That is why a PR was just merged to remove port 22 from php-fpm container. :) I think I was the one to submitted a PR that was merged that added that port. :( Still wiping the egg off of my face!!! :(

oprudkyi:
:+1:

PR #280, opened 2016-09-02
Ansible provision workspace

computerfr33k:
This PR will provision the workspace container using ansible to speed things up and reduce the image size. The growing list of software to install and build time is slowing things down and adding more intermediate images.

This PR requires the laradock/workspace container is be updated to 1.2
Related PR: https://github.com/LaraDock/workspace/pull/6


Mahmoudz:
@computerfr33k first this PR is awesome. 

It's definitely cleaner solution than what we currently have and a better alternative. However after re-thinking it I think using Ansible (an additional tool) will make some people uncomfortable submitting PR's as not everyone is familiar with it. And if we solve that problem we'll get into the debate of why Ansible not chef or puppet or...

Thanks again for the effort you put into making this project great.

lucafregoso:
@computerfr33k are you using this ansible provision in production?

computerfr33k:
I am not using Laradock in production.

PR #312, opened 2016-09-15
Implemented support for wkhtmltopdf.

Lednerb:
With this pull request, there will be support for wkhtmltopdf available.

It's just enabling the installation and (re)building the php-fpm container.

It uses the latest wkhtmltopdf version through @h4cc's repo [here](https://github.com/h4cc/wkhtmltopdf-amd64).

To generate pdfs the user has only to install @barryvdh [Laravel Snappy PDF Wrapper](https://github.com/barryvdh/laravel-snappy)


Lednerb:
@Mahmoudz maybe I should add this to the workspace container too...

I'm working with pdf generation via a laravel job / `queue:listen` and want to use it via the local `laradock` user so it has to run on the workspace container.

What do you think?

Mahmoudz:
Hi @Lednerb, 

Thanks for your contribution. The PR looks nice.
I won't be able to merge this since it's very case specific and is not gonna be used by many users. 

These kind of tools and modifications are better to be added on the user side, this way we keep LaraDock as simple as possible for people to modify it and use it.

I might have to remove some existing tools as well, and I will try to find solution for documenting these changes for other people to help them implement them easily (when I find the best solution to do this I will add your PR to the docs).

Thanks again.
Best,

PR #315, opened 2016-09-16
Windows Powershell command to enter container

mjcriswell:
Allows a Windows Powershell user to enter a running container to avoid the following error message.
Interactive mode is not yet supported on Windows.
Please pass the -d flag when using `docker-compose exec`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,12 @@ You can select your own combination of Containers form the list below:
 docker-compose exec workspace bash
 ```
 <br />
+<br>
+2a - Alternatively, Windows Powershell users can execute the following command to enter a running container.
+```bash
+docker exec -it {container-id} bash
+```
+<br />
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -269,6 +269,12 @@ You can select your own combination of Containers form the list below:
 docker-compose exec workspace bash
 ```
 <br />
+<br>
+2a - Alternatively, Windows Powershell users can execute the following command to enter a running container.
+```bash
+docker exec -it {container-id} bash
```
Mahmoudz: Why not using docker-compose? you can install Docker Compose on Windows!! and accessing the containers using Docker directly is also always an option.

korneliuskristianr:
Hi @mjcriswell , On windows you instead of using ```docker-compose```, type ```docker exec -it laradock_workspace_1 bash```

Issue #332, opened 2016-09-26
MySQL port (3306) exposed by default


philtrep:
According to https://github.com/docker/compose/issues/2260, we cannot yet override the ports options in docker-compose, could we remove the database ports by default so the `docker-compose.yml` is viable for production?

If you'd really need host access to the database you could just add a `docker-compose.override.yml` with the port mapping.

Mahmoudz:
It's possible but not sure if would be easy for the current users, to adapt that!

philtrep:
@Mahmoudz

From what I gather we could change the `ports` to `expose` without requiring most users to adapt. The proper way to execute any command is from the `workspace` anyhow.

Maybe it's just me but the default behavior should be restrictive rather than permissive, especially when it comes to database access.

PR #339, opened 2016-10-03
Remove port fowarding for MySQL and MariaDB

philtrep:
As discussed in #332, the database ports should not be forwarded by default since there is no way to override array values currently.

The dockerfiles for both the MySQL and MariaDB already expose the 3306 port so there is no need to do it again in the `docker-compose.yml`.

I only made the change for the MySQL and MariaDB since these are the ones I used, if this PR is considered I'll take the time to test all the other database containers for potential shortfalls.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -104,8 +104,6 @@ services:
         build: ./mysql
         volumes_from:
             - volumes_data
-        ports:
-            - "3306:3306"
         environment:
             MYSQL_DATABASE: homestead
             MYSQL_USER: homestead
@@ -131,8 +129,6 @@ services:
         build: ./mariadb
         volumes_from:
             - volumes_data
-        ports:
-            - "3306:3306"
         environment:
             MYSQL_DATABASE: homestead
             MYSQL_USER: homestead


Mahmoudz:
@philtrep first thanks for your contributions.

I don't think after removing this will remain clear for the new users that they can change the port from here, without digging into the container code.. Could you share what are the benefits of the removing them? 

Thanks again

philtrep:
@Mahmoudz you are welcomed! thank you for creating/maintaining LaraDock :)

Having the 3306 port forwarded by default.

It's explained in https://github.com/docker/compose/issues/2260 that we cannot yet override the `ports` (or any other array values for that matter). Therefore my only way to NOT have the 3306 port forward is to edit the `docker-compose.yml` file.

This is a problem for 2 reasons:
**1.**
In the [official docker-compose documentation](https://docs.docker.com/compose/production/) it says:
> you’ll probably want to define an additional Compose file, say production.yml, which specifies production-appropriate configuration
Since we cannot here use an additional Compose file to overwrite the `ports` value, we are going against the recommended method in the documentation.

**2.**
If you are programmatically grabbing the LaraDock files, the only way to replace the `ports` value is to do a in-file string replace. Currently, I have to do a `sed` replacement on the `docker-compose.yml` since my `docker-compose.override.yml` cannot override that value.

As far as the downsides go:
**1.**
I don't think that many users are remotely connecting to the database from outside the Compose network. Those who do (and _really_ want to) shouldn't have any problems in adding the extra line in their environment specifig Compose file.

**2.**
Security wise it makes more sense for this to be an opt-in instead of an opt-out. As far as I know, the current MySQL configuration doesn't use SSL by default (looking at my logs right now I get `[Warning] Failed to set up SSL because of the following SSL library error: SSL context is not usable without certificate and private key`) so we are providing a remotely accessible, unencrypted connection to a database with known default user/pass.

Mahmoudz:
Thanks for the clarification. Looking forward for more contributions from you :)

Mahmoudz:
Hi @philtrep since me and some other users started facing some problems with this (mainly because it wasn't documented) and in all cases it's an extra config that most of us will have to add manually, due to the fact that 90% of the users are using LaraDock on development only, I'm thinking of making the current `docker-compose.yml` more like the development version of the file (and add the ports forwarding back to it). 

And add a new section to the readme such as "LaraDock on production" where in this one we encourage developers to write their own `production-docker-compose.yml` and we advise them to only add the containers they want to that file and we tell them to not use the port forwarding there.. maybe we can include a basic files with some assumptions in it to make it even faster to create that file.

What do you think?

Mahmoudz:
@philtrep Kindly find these changes #352 and please let me know what you think of this approach?

philtrep:
Sounds good @Mahmoudz . There's definitely too much confusion about this. I started working on a PR but I see you've already done it :)

Mahmoudz:
@philtrep yep, you can never make one thing that suits everyone. Currently the priority is to make LaraDock work on development with the least effort possible (2 to 3 steps). I'm gonna make a video tutorial covering version the 4 very soon, the last video made was for v2.2 and many changes occurred since then..

philtrep:
@Mahmoudz gotcha, I'm gonna try to focus on issues related to initial setup.

PR #367, opened 2016-10-14
Fixed mysql tag dockerfile

aletundo:
Signed-off-by: Alessandro Tundo <alessandrotundo94@gmail.com>

### DIFF
--- a/mysql/Dockerfile
+++ b/mysql/Dockerfile
@@ -1,4 +1,4 @@
-FROM mysql:mysql:5.6
+FROM mysql:5.6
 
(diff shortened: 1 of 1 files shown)


philtrep:
@aletundo Good catch, thank you for your contribution :)

brianvice:
it may also be desirable to have mysql:latest which would provide mysql 5.7 that offers the JSON field type. [https://dev.mysql.com/doc/refman/5.7/en/json.html](https://dev.mysql.com/doc/refman/5.7/en/json.html)

philtrep:
@brianvice please refer to #364

Issue #366, opened 2016-10-14
Mysql downgrade to 5.6 replace mysql:mysql:5.6 to mysql:5.6


henrychavez:
@Mahmoudz I had a problem when I try to run the command 
`docker-compose up -d  nginx mysql`

This is the error 
`Step 1 : FROM mysql:mysql:5.6
ERROR: Service 'mysql' failed to build: Error parsing reference: "mysql:mysql:5.6" is not a valid repository/tag`

To solve the problem I had to replace 
`FROM mysql:mysql:5.6` to `FROM mysql:5.6`

philtrep:
this was solved by #367

PR #368, opened 2016-10-14
Update Dockerfile

brianvice:
Update to latest version of mysql. This will (at this time) provide 5.7.16 which will allow use of the json field type.

### DIFF
--- a/mysql/Dockerfile
+++ b/mysql/Dockerfile
@@ -1,4 +1,4 @@
-FROM mysql:5.6
+FROM mysql:latest
 
 MAINTAINER Mahmoud Zalt <mahmoud@zalt.me>
 


philtrep:
@brianvice Thank you for your contribution but please refer to #364 and #366

Mahmoudz:
@brianvice @philtrep lool I just made that 5.6 today, I was thinking of defining the versions in every dockerfile and then manually updating them after testing each release! Because this is not the first time users find bugs with the latest updates, and we cannot easily reproduce these errors since we all have our own containers.. 

Basically we have 2 options:
Option 1: keep everything to the latest and wait for bugs to occure and fix them.
Option 2: manually upgrade each file version and make sure all users have working versions.

I know each has it's own disadvantages, I personally don't mind any one. What do you guys think?

brianvice:
@Mahmoudz @philtrep 
Hey guys, thanks for getting back to me on this PR. I saw the update to 5.6. My reason for having 5.7 is that it allows for the full function of laravel's database migration features. With 5.6, if someone tries to use the json field type they'll get an SQL syntax error when performing their migrations.

While it's easy enough to just swap out the version number in the dockerfile and rebuild the container, I think the second option you mention is a good one. As in the php container, we could define multiple dockerfiles and then give the user the option of which image to use. But for completeness and in order to provide up to date packages I don't see a problem with this option.

philtrep:
@Mahmoudz @brianvice 
I think manually selecting the version is the best idea. Brian's idea of having multiple Dockerfile is fine by me BUT we should try to limit the amount of Dockerfile we have per image type so as not to have to maintain a dozen different Dockerfile for each image type.

The way I see it, we could focus on keeping the core Laravel containers like `php-fpm`and `mysql` up to date with the latest versions and keep the secondary/less used containers either on LTS or latest major release.

For example, we could support `mysql` 5.6 and 5.7 but for `mongo` only 3.2. That way we maximize ease of use for most users while keeping issues to a minimum.

Mahmoudz:
@brianvice @philtrep thanks for your replies. Phill I totally agree with you, w'll define the versions for the default most basic containers and let the rest use the :latest.

PR #370, opened 2016-10-15
Certbot container

philtrep:
I added a certbot container, it should be enough to generate a certificate and keep it updated. NGINX serves `/var/www/letsencrypt` under `/.well-known/acme-challenge`. The generated certificates are available on the host in `laradock/data/certbot/certs` and in the containers in `/var/certs/`.

**This is an initial PR** please test it and add your solutions for _apache_ and _caddy_.


g0ld3lux:
caddy has own engine for automation of the ssl certificate ive tested caddy made a tweak in caddyfile based on their documentation, i think the only thing left now is apache...

philtrep:
@g0ld3lux Good point.

Then we just need an Apache solution and a README entry on how to use the Certbot container. I'm pretty busy this week so if anyone wants to take a crack at it, go ahead.

philtrep:
We still need a Apache solution for this.

Mahmoudz:
@philtrep that PR is here for long :P I'm gonna merge it, and later we can add whatever missing like the docs.. Thanks,

CSNWEB:
Seems like this is still missing docs?

swash13:
can someone add documentation about how to setup certbot with nginx ?
Not only for digitalocean, but for amazon ec2 also.

xlstefan:
Still no documentation regarding nginx and certbot automatic renewal?

PR #384, opened 2016-10-23
phpstorm-debugging-guide

LarryEitel:
Guide to remote debugging Laravel in PHPStorm.


LarryEitel:
This is an initial crack at providing a guide to debugging Laravel in PHPStorm. Although it will be useful to anyone interested, it is not considered a release version. I felt it important to get SOMETHING on the table sooner than later. :) It would be great if other users of PHPStorm could follow this guide to confirm whether this guide is complete and accurate.

Mahmoudz:
@LarryEitel the documentation looks very nice. 

Just make sure to disable xdebug by default. And mention that xdebug needs to be enabled first..

Best,

LarryEitel:
@Mahmoudz I think it is pretty clean now. Don't hesitate to share any suggestions/corrections. 

One thing that could be done would be to squash commits. :)

One more thing that I find useful is to have a branch off of my fork which is all configured/wired up for debugging. Then all I have to do is merge from master as needed. Here is an example [LarryEitel/laradock/tree/phpstorm-debug](https://github.com/LarryEitel/laradock/tree/phpstorm-debug) for anyone interested.

Exadra37:
@LarryEitel no need to use a separated branch to debug if we go as per #424.

LarryEitel:
I agree with the idea to set/override all config values through .env. As for this debugging guide, I made a branch although it only consists of a guide. I made a branch on MY fork to set appropriate values to verify debugging works. This would not be necessary if these values could be read from .env. 

I DO wonder if docker-compose will behave well if it sees no change in the actual .yml file.

Exadra37:
Good point about docker compose detecting the changes... at least when I testing it by  run `docker-compose config` it was showing the changes to the `.env` file.

I will test later with `docker-compose up`.

philtrep:
@LarryEitel This looks really good. What do you think is missing for this to be considered a release version (aside from squashing commits)?

Even if the guide isn't 100% accurate, merging this in would at least give some tips to users and would increase the chances of others contributing.

LarryEitel:
Actually, I verified the debug guide works with current LaraDock. It doesn't affect any functionality. It only provides guidance to get debugging working. If/when anyone has a further question, I will try to update/adjust accordingly.

philtrep:
@LarryEitel Awesome, let me know when you're done squashing and I'll merge this in! 😃

LarryEitel:
@philtrep OK, good to go. :)

philtrep:
@LarryEitel All done. Thanks again for your contribution!

PR #386, opened 2016-10-24
update Workspace port to 2222

dilone:
i was having problems with the port 22, is already in use on my PC.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,7 +26,7 @@ services:
             # IMPORTANT: Replace with your Docker Host IP (will be appended to /etc/hosts)
             - "dockerhost:10.0.75.1"
         ports:
-           - "22:22"
+           - "2222:22"
         tty: true
 
 ### PHP-FPM Container #######################################


philtrep:
@dilone Thanks for your contribution.

Do you have a reasoning behind using the port `2222` instead?

If it's only because of conflict, I don't think we should use the alternate ssh port by default, `22` is standard and if we follow the config for our other images we should be using standard ports for forwarding.

That being said, if you've had an issue with conflicting ports maybe we can address it in the debugging section of the documentation?

dilone:
i have ssh server running on my Fedora, taking the port 22 by default, changing the map to the 2222 works great for me, all the command runs the same way, i just think is good addition to the project.

philtrep:
@Mahmoudz what do you think? `22` is development centric since if you ssh into a server the `22` port is probably already taken but I have seen configs using `2222`...

Mahmoudz:
@philtrep @dilone it makes sense, seems changing the default `22` would solve it but I'm not sure if it will cause troubles to the current users!! I'll leave it for you guys to decide.

philtrep:
Okay I'll merge this in since it does make sense that you might already have the port `22` used by ssh on your host machine, we'll revise if we get too many issues about this.

Thanks again for your contribution @dilone :)

Mahmoudz:
should we mention that somewhere in the docs!?

philtrep:
@Mahmoudz good point! I haven't seen any mention of actual `ssh` (only `exec`) in the doc but we should indeed add a mention.

Mahmoudz:
@philtrep thanksss

Issue #415, opened 2016-11-08
Installing Yarn


jschaftenaar:
```
Step 46 : RUN if [ ${INSTALL_YARN} = true ]; then     curl -o- -L https://yarnpkg.com/install.sh | bash &&     echo "" >> ~/.bashrc &&     echo 'export PATH="$HOME/.yarn/bin:$PATH"' >> ~/.bashrc ;fi
 ---> Running in f5958a5ab3e3
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    94  100    94    0     0    226      0 --:--:-- --:--:-- --:--:--   225
100  3531  100  3531    0     0   4667      0 --:--:-- --:--:-- --:--:--  4667
Installing Yarn!
> Downloading tarball...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    92  100    92    0     0    199      0 --:--:-- --:--:-- --:--:--   199
100   595    0   595    0     0    637      0 --:--:-- --:--:-- --:--:--   637
100 2743k  100 2743k    0     0  1078k      0  0:00:02  0:00:02 --:--:-- 2505k
> Extracting to ~/.yarn...
> Adding to $PATH...
> We've added the following to your /home/laradock/.profile
> If this isn't the profile of your current shell then please add the following to your correct profile:

export PATH="$HOME/.yarn/bin:$PATH"

> Yarn was installed, but doesn't seem to be working :(.
ERROR: Service 'workspace' failed to build: The command '/bin/sh -c if [ ${INSTALL_YARN} = true ]; then     curl -o- -L https://yarnpkg.com/install.sh | bash &&     echo "" >> ~/.bashrc &&     echo 'export PATH="$HOME/.yarn/bin:$PATH"' >> ~/.bashrc ;fi' returned a non-zero code: 1
```
Yarn does not seem to install. Also looking in /home/laradock/.yarn/bin or /root/.yarn/bin no yarn is to be found. While it seems to be downloading to ~/.yarn. 

Any suggestions are appreciated.

moxar:
I just ran into the same issue.

moxar:
It seems the problem is that the $PATH of the workspace container cannot find node.

Here is a fix.

https://github.com/laradock/laradock/pull/418

Mahmoudz:
[...]

PR #418, opened 2016-11-09
Fix a bug that prevents yarn from installing #415

moxar:


Mahmoudz:
@moxar Thank you for your contribution.

PR #434, opened 2016-11-16
Fixes for nginx and mysql

dlnsk:
It seems more reasonably don't rebuild image every time when new application added, so just link its configurations as volume.

It very strange to downgrade mysql version from latest to 5.6 because it broke available databases. I offer to kill ONLY_FULL_GROUP_BY evil and to use 5.7 ;)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -75,6 +75,7 @@ services:
             - applications
         volumes:
             - ./logs/nginx/:/var/log/nginx
+            - ./nginx/sites/:/etc/nginx/sites-available
         ports:
             - "80:80"
             - "443:443"
--- a/mysql/Dockerfile
+++ b/mysql/Dockerfile
@@ -1,4 +1,4 @@
-FROM mysql:5.6
+FROM mysql:5.7
 
 MAINTAINER Mahmoud Zalt <mahmoud@zalt.me>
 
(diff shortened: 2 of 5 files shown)


philtrep:
@dlnsk Could you give a bit more context to:
`STRICT_TRANS_TABLES`
`NO_ZERO_IN_DATE`
`NO_ZERO_DATE`
`ERROR_FOR_DIVISION_BY_ZERO`
`NO_AUTO_CREATE_USER`
`NO_ENGINE_SUBSTITUTION`
Why are these required?

dlnsk:
@philtrep This is default set of keys for 5.7 but exclude ONLY_FULL_GROUP_BY

> The default SQL mode in MySQL 5.7 includes these modes: ONLY_FULL_GROUP_BY, STRICT_TRANS_TABLES, NO_ZERO_IN_DATE, NO_ZERO_DATE, ERROR_FOR_DIVISION_BY_ZERO, NO_AUTO_CREATE_USER, and NO_ENGINE_SUBSTITUTION.

[Manual](http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-setting)

philtrep:
@dlnsk Awesome! Thanks for your work.

PR #474, opened 2016-12-09
Fix load of mongodb library

AliveCorpse:
Fix warning message in the workspace when use mongodb library
```
PHP Warning: PHP Startup: Unable to load dynamic library '/usr/lib/php/20151012/mongodb.so' - /usr/lib/php/20151012/mongodb.so: undefined symbol: php_json_serializable_ce in Unknown on line 0
```


philtrep:
@AliveCorpse Thank you for you contribution!

PR #478, opened 2016-12-12
fix workspace load mongodb.so problem

lialosiu:
```
PHP Warning: PHP Startup: Unable to load dynamic library '/usr/lib/php/20151012/mongodb.so' - /usr/lib/php/20151012/mongodb.so: undefined symbol: php_json_serializable_ce in Unknown on line 0
```

Fix php warning when workspace set `INSTALL_MONGO=true` and mongodb.so can not be load.


lialosiu:
oh, I see a PR #474 already opened to fix this problem... forget me :)

PR #486, opened 2016-12-15
Fix mysql strict NO_ZERO_DATE

dlnsk:
NO_ZERO_DATE restrict to use 0000-00-00 00:00:00 as a date value, but Laravel use it for create_at, update_at etc. This mode don't allow to do any alter for mysql tables with zero values as default for columns.

Manual: http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_zero_date


philtrep:
@dlnsk Thanks!

PR #487, opened 2016-12-15
Laradock CLI Helper allow much simple use command line tools like npm…

dlnsk:
Laradock CLI Helper allow much simple use command line tools (from container) like npm, composer, gulp, bower etc. with your projects.

You need to use multisites configuration like (structure inside container after linking):
/var/www/project1
/var/www/project2
...

**Step 1** (install soft link in PATH folder):
cd laradock
sudo ldk install

**Step 2** (using):
cd ~/MyProjects/project1
ldk composer install   // Use composer to install dependencies
ldk composer <any_composer_command>
ldk artisan <any_artisan_command>
ldk npm <any_command>
ldk bower <any_command>
ldk gulp <any_command>

PS: You can rename script as you like.

### DIFF
--- a/ldk
+++ b/ldk
@@ -8,6 +8,25 @@
 # Version: 0.1.0
 #####################################################
 
+command=$1
+
+
+case $command in
+    init)
+        echo "Checking latest release..."
+        if [ ! -d "${PWD}/.laradock" ];
+        then
+            mkdir .laradock
+        fi
+        release_url=`wget -qO- https://api.github.com/repos/laradock/laradock/tags | sed -n '/tarball_url/p' | head -n 1 | sed -e 's/.*tarball_url.*\(http.*\)".*/\1/'`
+        echo "Using Laradock ${release_url##*/} release."
+
+        echo "Downloading..."
+        wget -qO- $release_url | tar xz --directory=${PWD}/.laradock --strip-components=1
+        echo "Done."
+        exit 0
+esac
+
 if [ -d "${PWD}/.laradock" ];
 then
     script_dir="${PWD}/.laradock"
@@ -17,6 +36,7 @@ else
         script_dir="${PWD}/../.laradock"
     else
         echo "Warning! Laradock not found for this project."
+        echo "Maybe you want to use 'ldk init' to download and install Laradock?"
         exit 1
     fi
 fi
--- a/ldk
+++ b/ldk
@@ -1,89 +1,24 @@
 #!/bin/bash
 
-command=$1
-params=${@:2}
-cur_dir_name=${PWD##*/}
-project_dir="/var/www/$cur_dir_name"
-script=$(readlink -f "$0")
-script_dir=$(dirname $script)
+#####################################################
+# This is a starter for Laradock
+# 
+# See details on https://github.com/laradock/laradock
+# Author: Dmitry Pupinin
+# Version: 0.1.0
+#####################################################
 
-node_init="source \"\$NVM_DIR/nvm.sh\" && export PATH=\$PATH:/home/laradock/.nvm/versions/node/\$(echo \"nvm current\")/bin"
-
-help_text="Laradock CLI v0.1
-The easiest way to use tools inside 'workspace' container.
-
-Tools available:
-    artisan
-    composer
-    node
-    npm
-    nvm
-    gulp
-    bower
-    install (makes this script global)
-    run (executes shell commands)
-
-Examples (run in project's directory on the host):
-    ldk artisan make:model MyNewModel
-    ldk composer install
-    ldk run php -v
-"
-
-
-case $command in
-    install)
-        echo "Installing..."
-        if [ "$EUID" -ne 0 ]
-        then
-            echo "Failed! Be shure that you run install with 'sudo'"
-            exit 1
-        else
-            read -p "Please, choose name of symlink [ldk]: " link_name
-            if [ "$link_name" == "" ]
-            then
-                link_name="ldk"
-            fi
-            lnk="/usr/bin/$link_name"
-            if [ -L $lnk ]
-            then
-                rm $lnk
-            fi
-            ln -s `realpath $0` $lnk
-            echo "Success! Symbolic link ${lnk} created."
-            echo "Now you can go in project directory and run '$link_name artisan'"
-            exit 0
-        fi ;;
-
-    run)
-        tail="$params" ;;
-
-    artisan)
-        tail="cd ${project_dir} && ./artisan $params" ;;
-
-    composer)
-        tail="${*} --working-dir ${project_dir}" ;;
-
-    node)
-        tail="$node_init && node" ;;
-
-    npm)
-        tail="$node_init && npm --prefix=${project_dir} $params" ;;
-
-    nvm)
-        tail="$node_init && nvm $params" ;;
-
-    gulp)
-        tail="$node_init && gulp $params --cwd ${project_dir}" ;;
-
-    bower)
-        tail="$node_init && bower $params --config.cwd=${project_dir}" ;;
-
-    *)
-        echo "$help_text" ;;
-
-esac
-
-if [ "$tail" != "" ]
+if [ -d "${PWD}/.laradock" ];
 then
-    docker-compose -f $script_dir/docker-compose.yml exec --user=laradock workspace /bin/bash -c "$tail"
+    script_dir="${PWD}/.laradock"
+else
+    if [ -d "${PWD}/../.laradock" ];
+    then
+        script_dir="${PWD}/../.laradock"
+    else
+        echo "Warning! Laradock not found for this project."
+        exit 1
+    fi
 fi
+
+$script_dir/ldk-cli $*
--- a/ldk
+++ b/ldk
@@ -1,0 +1,89 @@
+#!/bin/bash
+
+command=$1
+params=${@:2}
+cur_dir_name=${PWD##*/}
+project_dir="/var/www/$cur_dir_name"
+script=$(readlink -f "$0")
+script_dir=$(dirname $script)
+
+node_init="source \"\$NVM_DIR/nvm.sh\" && export PATH=\$PATH:/home/laradock/.nvm/versions/node/\$(echo \"nvm current\")/bin"
+
+help_text="Laradock CLI v0.1
+The easiest way to use tools inside 'workspace' container.
+
+Tools available:
+    artisan
+    composer
+    node
+    npm
+    nvm
+    gulp
+    bower
+    install (makes this script global)
+    run (executes shell commands)
+
+Examples (run in project's directory on the host):
+    ldk artisan make:model MyNewModel
+    ldk composer install
+    ldk run php -v
+"
+
+
+case $command in
+    install)
+        echo "Installing..."
+        if [ "$EUID" -ne 0 ]
+        then
+            echo "Failed! Be shure that you run install with 'sudo'"
+            exit 1
+        else
+            read -p "Please, choose name of symlink [ldk]: " link_name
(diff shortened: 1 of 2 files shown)


ldk
```diff
@@ -0,0 +1,46 @@
...
+        tail="gulp $params --cwd ${project_dir}" ;;
+
+    bower)
+        tail="bower $params --config.cwd=${project_dir}" ;;
+
+esac
+
+if [ "$tail" != "" ]
+then
+    docker-compose -f ~/Projects/laradock/docker-compose.yml exec --user=laradock workspace /bin/bash -c "$tail"
```
philtrep: Instead of `~/Projects/laradock/docker-compose.yml` this should be fetching the `docker-compose.yml` file based on the script's location. Maybe allow specifying the `docker-compose` config file but keep `docker-compose.yml` as default?

docker-compose.yml
```diff
@@ -27,7 +27,7 @@ services:
                 - INSTALL_WORKSPACE_SSH=false
                 - PUID=1000
                 - PGID=1000
-                - NODE_VERSION=stable
+                - NODE_VERSION=6
```
philtrep: I don't think we should lock the node version.

philtrep:
@dlnsk You are on a roll! This is definitely interesting.

@appleboy @Mahmoudz I think we should discuss this since its a brand new feature.

Adding a CLI to run bash commands in the workspace container has obvious benefits but I do wonder about:

1. Could using bash prevent Windows users from using this? Go and Python could be viable alternatives.
2. Could adding the symlink could be leading to confusion if using multiple instances of LaraDock?
3. What do the Node changes have to do with the CLI? Is there a problem with using the current installation?

@logaretm 's (CLI)[https://github.com/logaretm/laradock-cli] implements management commands, it would be nice if the building blocks for the CLI were to allow for the inclusion of similar commands in the future.

What do you guys think?

workspace/Dockerfile
```diff
@@ -148,39 +148,23 @@ USER laradock
...
-# Wouldn't execute when added to the RUN statement in the above block
-# Source NVM when loading bash since ~/.profile isn't loaded on non-login shell
-RUN if [ ${INSTALL_NODE} = true ]; then \
-    echo "" >> ~/.bashrc && \
-    echo 'export NVM_DIR="$HOME/.nvm"' >> ~/.bashrc && \
-    echo '[ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh"  # This loads nvm' >> ~/.bashrc \
-;fi
-
-# Add NVM binaries to root's .bashrc
 USER root
```
appleboy: Why remove nvm?

docker-compose.yml
appleboy: agree with @philtrep

appleboy (changes requested):

docker-compose.yml
dlnsk: Not always new major version fully backward compatible (see on modes of mysql 5.7)
It not very good if with enabling xdebug I bring new version of node and something crash...

workspace/Dockerfile
dlnsk: Just very quick method to force node works. ;о))

ldk
dlnsk: It my hardcoded fault... sorry!

docker-compose.yml
philtrep: I tend to agree with you that it is far preferable in the long-term to lock the node version but for new users I think its better to have something that offers the latest version upon which they can build. Most LaraDock users end up modifying the `docker-compose` file anyhow.

We've discussed this previously in #368.

workspace/Dockerfile
philtrep: It does make building the container faster but I think we should keep `nvm for the time being`.

ldk
philtrep: No problem

logaretm:
@philtrep I kinda stopped maintaining it since the v 4.0.0 because it was much easier to setup multiple apps, however We can try to transfer ownership if you are interested to take over or I can help you with implementing better CLI tool since I'm not very skilled at it.

docker-compose.yml
Mahmoudz: Well obviously each way has it's own disadvantages.. I think we should make this clear in the docs and I agree with @philtrep let's keep this stable until it breaks!

ldk
dlnsk: Fixed

dlnsk:
Now asking the name of symlink, so you can install script for multiply instances of Laradock

appleboy:
@philtrep @Mahmoudz  We can create new cli repo to implement feature like [laradock-cli](https://github.com/logaretm/laradock-cli) using [Golang](https://golang.org/) for first version which support `linux`, `windows` and `darwin` OS. I will take it.

workspace/Dockerfile
dlnsk: I am not skilled in node and nvm.
It used strange path and I had problem with running node, but I'll try once again with older version of image.

appleboy: Please show detail error log, everybody can help to resolve node issue.

dlnsk:
I did reverted to original node install with nvm. It was not simple... :o)
But I don't fully understand why nvm needs. Just to install _any_ version of node? Is it really need?
Main concept as I understand is switching between number of versions. In this case I suggest to add in Dockerfile availability to install many versions of node in same time.

philtrep:
@dlnsk @appleboy Okay, so we can put all the project management commands in the separate repository and keep the container internal command aliases like in this PR within LaraDock?

@logaretm Like @appleboy pointed out we can start a new repository, your contribution will be more than welcomed :)

@dlnsk That's a good point about `nvm`. I think this has to do with making sure that node is installed properly on the container and that you can switch version without rebuilding. Also it might be getting used if you have multiple projects within a same LaraDock instal. Could you please squash the commits for this PR? I will merge when you are done!

appleboy:
@philtrep We can squash and merge this commit automatically. 😄 

screen shot 2016-12-20 at 9 19 28 am

ldk
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+command=$1
+params=${@:2}
+cur_dir_name=${PWD##*/}
+project_dir="/var/www/$cur_dir_name"
+script=$(readlink -f "$0")
+script_dir=$(dirname $script)
+
+node_init="source \"\$NVM_DIR/nvm.sh\" && export PATH=\$PATH:/home/laradock/.nvm/versions/node/\$(echo \"nvm current\")/bin"
```
appleboy:
```
if [ -f "${HOME}/.nvm/nvm.sh" ]; then
    export NVM_DIR=$HOME/.nvm
    source ${HOME}/.nvm/nvm.sh > /dev/null 2>&1
    nvm use stable > /dev/null 2>&1
fi
```

ldk
```diff
@@ -0,0 +1,63 @@
...
+        fi ;;
+
+    artisan)
+        tail="${project_dir}/artisan $params" ;;
+
+    composer)
+        tail="${*} --working-dir ${project_dir}" ;;
+
+    node)
+        tail="$node_init && node" ;;
```
appleboy: remove $node_init

ldk
```diff
@@ -0,0 +1,63 @@
...
+
+    npm)
+        tail="$node_init && npm --prefix=${project_dir} $params" ;;
```
appleboy: Same as above

ldk
```diff
@@ -0,0 +1,63 @@
...
+
+    nvm)
+        tail="$node_init && nvm $params" ;;
```
appleboy: Same as above

ldk
```diff
@@ -0,0 +1,63 @@
...
+
+    gulp)
+        tail="$node_init && gulp $params --cwd ${project_dir}" ;;
```
appleboy: Same as above

ldk
```diff
@@ -0,0 +1,63 @@
...
+
+    bower)
+        tail="$node_init && bower $params --config.cwd=${project_dir}" ;;
```
appleboy: Same as above

appleboy (changes requested):

ldk
dlnsk: @appleboy No, no! This script runing on host but nvm installed into container!
node_init should execute inside container, without it you can't use node at all!

appleboy: OK, Got it. You can use the following script in your `node_init ` variable.

```
export NVM_DIR=$HOME/.nvm
source ${HOME}/.nvm/nvm.sh > /dev/null 2>&1
nvm use stable > /dev/null 2>&1
```

dlnsk: This is not need because NVM_DIR and anything else is already did by Dockerfile. In additional $HOME/.nvm is wrong for root user.

philtrep:
@appleboy I know, just wanted to give @dlnsk some time to cleanup 😃

dlnsk:
I have an idea how we can use ldk for many installation without different links. We should make ldk_starter script (or maybe just in ldk) that will be installed and uses working directory to check laradock in one of three paths:
./laradock
../laradock
<some global>/laradock

appleboy:
Could we move to https://github.com/laradock/cli to discussion what features we need?

dlnsk:
Let's move.

appleboy:
ref: https://github.com/laradock/cli/issues/1

dlnsk:
Squashed.

philtrep:
@dlnsk I didn't think about it before but would you mind adding some usage documentation?

philtrep:
@dlnsk Good idea.

dlnsk:
I am not good in English, so check it please and correct... ;o)

ldk
```diff
@@ -0,0 +1,89 @@
...
+    composer
+    node
+    npm
+    nvm
+    gulp
+    bower
+    install (makes this script global)
+    run (runs any available system command)
+
+Examples (do it inside project's directory):
```
philtrep: Examples (run in project's directory on the host):

philtrep:
@dlnsk I added some comments. I think that some information and short summary should be available in the docs (`README`) as well.

philtrep:
@dlnsk When you have some time could you apply the last few fixes? I look forward to merging and starting to use this 😃

dlnsk:
@philtrep I haven't any uncommited features. Maybe we talk about something but I forgot?

philtrep:
@dlnsk Wow sorry this notification went by right under my nose. @appleboy has several comments to remove `$node_init` from the commands.

dlnsk:
Added laradock's starter. Now you can use per project or multi project configuration.
Not tested well.

Olivia101:
@dlnsk Will this work on Windows as well?

dlnsk:
@Olivia101 I don't know. Maybe with something like:
- https://msdn.microsoft.com/en-us/commandline/wsl/about
- http://win-bash.sourceforge.net/
- http://www.cygwin.com/

dlnsk:
With `ldk init` allowing to download latest laradock release into current project's directory.

dlnsk:
Updated to current 'master'

zuohuadong:
@appleboy  
In windows, we can use "cygwin".

https://www.cygwin.com

But you have to use Docker.exe

Mahmoudz (approved):

pratikid (approved):
LGTM

dlnsk:
Instead of script I made just alias:
https://gist.github.com/dlnsk/83dc033129a19e17c5c3208c7fc02f95

Mahmoudz:
Closing as stale — no activity in 4+ years. The codebase has changed significantly since this was opened. If still relevant, please rebase against current master and open a fresh PR. Thanks for the original contribution.

PR #506, opened 2016-12-22
Add `art` alias for `php artisan` - like homestead

Jono20201:
Since moving from Homestead to Laradock this is definitely something that I have missed, so I've added it to the Dockerfile.


philtrep:
@Jono20201 Thanks!

PR #528, opened 2017-01-05
pull from master

tristanbailey:


tristanbailey:
stupid me today cant use git

PR #538, opened 2017-01-10
Updated Crontab for Workspace

### LINKED ISSUE #459: Crontab Not Starting On Run
As discussed in #14, The crontab is configured in the workspace container but does not start until it is manually edited via `crontab -e`. Seems to be an issue with installing and activating the cron.

When editing via `contab -e` no changes are required, just open and save. The standard Laravel cron task immediately starts running correctly.

I've experienced this issue on three different projects both on Ubuntu Server and OSX.

sixlive:
- Crontab now gets added to `/etc/cron.d`
- Base crontab now runs as the `laradock` user

This update resolves #459.

### DIFF
--- a/workspace/Dockerfile
+++ b/workspace/Dockerfile
@@ -79,7 +79,7 @@ RUN if [ ${COMPOSER_GLOBAL_INSTALL} = true ]; then \
 #####################################
 USER root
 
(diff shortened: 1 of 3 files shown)


philtrep:
@sixlive Thank you for your contribution!

PR #555, opened 2017-01-16
Fix Workspace Cron Permissions

sixlive:
When I submitted #538 it looks like I overlooked the crontab file
permissions for the Laradock user. This change ensure that all
crons added during build are set to the correct permission level.


philtrep:
@sixlive Thank you

PR #602, opened 2017-02-10
Added queue worker to workspace as command

leu2tm:


leu2tm:
When using some sort of automatic deployment, it's kinda cumbersome to navigate into the container and then start the process through a script.

Also when just trying to restart/upping the container again, you'll have to repeat the procedure.

scoutman57:
As this is project specific, why don't you create a bash script or just maintain this in your file? As this does not affect everyone (people not using worker queues). Just my thoughts.

leu2tm:
Until now I have maintained my own file, but I thought this would be great for people who uses workers, and even for people that doesn't use it yet.

But if you feel like it you may close this PR.

philtrep:
@leu2tm First off, thank you for the PR 😃.

I agree with @scoutman57 , this seems a bit too project specific to be done in this way.

Could we make this a build argument instead? People who want to do the same would just have to set `RUN_ARTISAN_QUEUE` to true.

Just write a little doc to go along with it and I think we're golden.

leu2tm:
@philtrep sounds reasonable, i'll look into it :)

scoutman57:
To piggy back off what @philtrep  stated. 

I think it could just be a `RUN_ARTISAN` to true or something similar that will execute different things that you need like migrations or something. This could even be extended to composer and npm.

leu2tm:
@scoutman57 are you suggesting that ```RUN_ARTISAN``` should take in a string, so it can run artisan commands?

scoutman57:
that or an array of params. Not sure which would work best since some artisan command require additional params such as `artisan migrate --seed`

leu2tm:
@scoutman57 Hmm I think it would be best to scope it queue first, since some might use migration or other custom commands in a certain order.

leu2tm:
I started using ```docker-compose run -d workspace php artisan queue:work``` as it seems like a better way to start queue worker.

PR #604, opened 2017-02-10
Safer opcache settings

gagarine:
Some proposed changes:
- Drastically augment the max accelerated files. Drupal has easily 10000+ php files, same for symphony and I guess for larevel.
- remove fast shutdown, it's still create segfault in some situation and I didn't find benchmark that say "use this!".
- validate files timestamp, it's safer for dev and newbe. Advance user can set it to 0 for production
- validate_freq to 0 so validate only once by request.


philtrep:
@gagarine Awesome, thanks!

PR #617, opened 2017-02-19
don't load opcache extension twice

alchimik:


alchimik:
The extension is loaded in  /usr/local/etc/php/conf.d/docker-php-ext-opcache.ini (if INSTALL_OPCACHE=true). 
see https://github.com/laradock/laradock/blob/master/php-fpm/Dockerfile-70#L171
no need to load twice

> Does this work with CLI version too?

yes


> Why if you are going to move the extension out why not the other settings?
> Well the setting from php-fpm/opcache.ini be read into the in from the docker-php?

i do not understand

alchimik:
up

PR #632, opened 2017-02-23
xdebug.ini typo fix

flashbag:
In the https://github.com/laradock/laradock/blob/master/php-fpm/Dockerfile-70#L75
and also in https://github.com/laradock/laradock/blob/master/php-fpm/Dockerfile-56#L63
there is one name for the xdebug configuration file -> **xdebug.ini**

But in the script that controlling xdebug, there is another:
https://github.com/laradock/laradock/blob/master/xdebugPhpFpm#L39
https://github.com/laradock/laradock/blob/master/xdebugPhpFpm#L61


alchimik:
Wrong

```
root@xxx:/# cat /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini
zend_extension=/usr/local/lib/php/extensions/no-debug-non-zts-20151012/xdebug.so
```
current path is correct

file  /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini created by `docker-php-ext-enable xdebug` command
for example here - https://github.com/laradock/laradock/blob/master/php-fpm/Dockerfile-56#L59

PR #637, opened 2017-02-25
renamed docker-compose.yml to docker-compose.template.yml

Treggats:
let people define their own compose config, but keep the original as an
example


philtrep:
@Treggats Thanks for the PR. I don't think this is the right way to go to have people build their own composefile. The reason being that most new users shouldn't need to create their own compose file and their initial setup should be as easy as possible. My suggestion is to instead create documentation explaining how they can create their file for their specific needs.

Treggats:
Seems reasonable, although in my opinion the current compose config is rather big. Even new users usually don't everything Laradock has to offer. My approach would be to start with a webserver, php-fpm and a database. Maybe using a script to create the compose config file.

Olivia101:
@Treggats I agree. I am planning to use LaraDock to create a dev environment for building WordPress themes. I would still like to have xDebug, composer, phpMyAdmin, memcached and elasticsearch out of the box.

philtrep:
@Treggats @Olivia101 For the moment your best option is to build the compose file yourself. A script/cli to create configurations on-the-fly is a great idea, we're exploring the idea on https://github.com/laradock/cli/issues/1 . You are more than welcomed to participate! :)

Treggats:
Will do, looks like an interesting read

PR #642, opened 2017-02-27
Multiple Projects support improvement

Mvzundert:
Moved the volumes setup to a .env file so you can have all projects in the folder above laradock mapped to var/www/ in your container.

This still means you can setup individually by mapping it differently in the .env file but does give you more flexibility.

Still working on the other containers.


Mahmoudz:
@Mvzundert this looks nice, please ping me when you're done.

Mvzundert:
@Mahmoudz I think this should be mostly done (for now) I'd like to split up integrating other stuff like Wordpress, Symfony, Drupal etc into different branches for clarity.

What I've done here:
- Introduced .env file (with env-example) to enable you to set most options in docker-compose.yml from the env file.
- Made the application container accept multiple folders (so no more mapping  - ../sample/:/var/www/sample or stuff like ../:/var/www which was also moved to the .env file.
- Split up mysqli and the tokenizer into seperate options (having them under code igniter was weird since there are more things (wordpress for example) that might need mysqli and it feels very strange having and hard to locate when it's under code igniter.
- Apache docker file now also maps to /var/www instead of /var/www/public (again to support more options and not have to default to creating public folders)
- Removed all homestead references instead I used "default"  (in code not in the readme's etc)

Mahmoudz:
@Mvzundert I've create a separate [branch](https://github.com/laradock/laradock/tree/multiple-projects-improvement) for this, made some updates as well. Should be ready after updating the docs, and performing some tests.

PR #650, opened 2017-03-03
Add mailhog for mail debugging

wayanjimmy:
Hi I'm try to help adding mailhog for mail debugging

PR #651, opened 2017-03-03
Add MailHog container

dtunes:


alchimik:
please add [ssmtp](https://github.com/gagarine/docker-php/blob/master/Dockerfiles/phpfpm/7.0/Dockerfile#L24) or equivalent (i use msmtp)
`mailhog/Dockerfile` file is not needed. Replace `build: ./mailhog` -> `image: mailhog/mailhog`

dtunes:
@alchimik I feel like ssmtp should be its own container...

alchimik:
if ssmtp is in a separate container, how php will use it?
https://github.com/gagarine/docker-php/blob/master/Dockerfiles/phpfpm/7.0/config/etc/php/7.0/fpm/php.ini#L1022

dtunes:
Sorry, so you want to just forward all emails to an external SMTP host?

alchimik:
to mailhog SMTP host - https://github.com/gagarine/docker-php/blob/master/Dockerfiles/phpfpm/7.0/config/etc/ssmtp/ssmtp.conf#L10

I want to be able to use the `mail` function, forward all email sended by `mail` function to mailhog.

dtunes:
Not sure how putting ssmtp on this container would allow php-fpm container to use it via the mail function directly

alchimik:
I meant to install ssmtp in php-fpm, not in mailhog container

Plus do the installation ssmtp optional. Something like:
```
COPY ./ssmtp.conf /etc/ssmtp/ssmtp.conf
COPY ./php_sendmail.conf /usr/local/etc/php/conf.d/sendmail.conf

ARG FORWARD_MAIL_TO_MAILHOG=false
RUN if [ ${FORWARD_MAIL_TO_MAILHOG} = true ]; then \
    apt-get -y update && \
    apt-get -y ssmtp && \	
;fi

RUN if [ ${FORWARD_MAIL_TO_MAILHOG} = false ]; then \
    rm /etc/ssmtp/ssmtp.conf  && rm /usr/local/etc/php/conf.d/sendmail.conf \
;fi
```

ssmtp.conf
```
mailhub=mail:1025
hostname=web
```

php_sendmail.conf
```
; Send to mailhog
sendmail_path = "/usr/sbin/ssmtp -t"
```

**Just an example, not tested**

dtunes:
I see, it was really confusing because you added this comment in the MailHog PR. Maybe just add that as a separate request? I'm sure it's an easy task, if I have time I might do another PR for it, but for now, just add that as a separate request suggestion?

dtunes:
FYI, I've added the Dockerfile because its recommended in the docs, I guess it just makes it easy for configuring the container later.

alchimik:
> I see, it was really confusing because you added this comment in the MailHog PR. Maybe just add that as a separate request? I'm sure it's an easy task, if I have time I might do another PR for it, but for now, just add that as a separate request suggestion?

Yes, you're right

little question - why do we need `links: - php-fpm`?

dtunes:
so that you can reference the mailhog container from php-fpm e.g. MAIL_HOST=mailhog

alchimik:
> so that you can reference the mailhog container from php-fpm e.g. MAIL_HOST=mailhog

It works without links - [Networking in Compose](https://docs.docker.com/compose/networking/)
> By default Compose sets up a single network for your app. Each container for a service joins the default network and is both reachable by other containers on that network, and discoverable by them at a hostname identical to the container name.

dtunes:
Oh true, I haven't heard the v2 changes yet...

philtrep:
@dtunes This is a duplicate of #650. Please contribute to the initial PR instead of creating a new one.

dtunes:
Yeah, I literally wrote that one night, and published the other night, only
after I saw that PR. Just approve whichever and close the second one.

PR #653, opened 2017-03-05
Add minio container

thorerik:
Minio is an s3 compatible server written in go, so you can test and run using s3-like storage rather than local fs.


alchimik:
1. imho - very case specific and is not gonna be used by many users
2. something wrong with `docs/index.html`

thorerik:
2. Curious, reverted it, I didn't touch the index.md in the root, and I'm not well enough faired with hugo to know what caused that.

1. same could be said for selenium and some other tools in the project, that said, it being available for users to learn how S3 works without needing to touch AWS can't be bad, I intend to use it, and will maintain the Dockerfile (not that it's super complex anyways), and if others can benefit from it as well, why not?

Mahmoudz:
@thorerik looks awesome! I'm gonna consider switching to jekyll

PR #654, opened 2017-03-06
workaround for 'Xdebug breaks on access to class static property'

alchimik:
see https://github.com/docker-library/php/issues/133


philtrep:
@alchimik Thanks for the PR!

PR #655, opened 2017-03-06
fix selenium - mount /dev/shm

alchimik:
see https://github.com/SeleniumHQ/docker-selenium#running-the-images
> When executing docker run for an image with chrome browser please add volume mount -v /dev/shm:/dev/shm to use the host's shared memory.


philtrep:
@alchimik Thanks!

PR #670, opened 2017-03-14
Mssql support

cabrerabywaters:
Adds Support for MSSQL in php7.0-fpm container


Mahmoudz:
@cabrerabywaters also conflict here, please solve it to merge. Thanks

cabrerabywaters:
@Mahmoudz All set here...(Y)

thorerik:
May I suggest bundling in the [mssql-server-linux](https://hub.docker.com/r/microsoft/mssql-server-linux/) docker image so that you can have a fully self-contained dev environment, also with MSSQL.

cabrerabywaters:
@thorerik I use that container too but...Is there any way to integrate it with PHP-FPM that runs in this container? This is not for running a MSSQL DB, but to use PHP SQL extensions.

Mahmoudz:
@cabrerabywaters can you a MSSQL container in Laradock, to give us more control from one place. 
See any other DB support and try to do the same, you can use the image mentioned by @thorerik as base image.

cabrerabywaters:
@Mahmoudz Sure! let me take a look at the other DBs and I'll see what i can do ;)

Mahmoudz:
@cabrerabywaters something simple to look at is [NEO4J](https://github.com/laradock/laradock/blob/master/neo4j/Dockerfile). Best,

PR #673, opened 2017-03-14
Php worker

cabrerabywaters:
Laravel queue work container to listen to app's queues.


cjmaxik:
Nice one, really more reliable than my variant http://laradock.io/documentation/#run-laravel-queue-worker

docker-compose.yml
```diff
@@ -74,6 +74,15 @@ services:
         environment:
             # IMPORTANT: Set the Remote Interpreter entry matching name to `laravel`
             - PHP_IDE_CONFIG=serverName=laravel
+            
+### PHP Worker Container #################################
+    php-worker:
+        build:
+            context: ./php-worker
+        volumes:
```
cjmaxik: Should be relative to `applications` container:

```yaml
        volumes_from:
            - applications
```

php-worker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+
+#
+#--------------------------------------------------------------------------
+# Final Touch
+#--------------------------------------------------------------------------
+#
+
+RUN rm -r /var/lib/apt/lists/*
+WORKDIR /etc/supervisor/conf.d/supervisord.conf
```
cjmaxik: Should be at least `WORKDIR /etc/supervisor/conf.d/` since existed variant is not working:
```
ERROR: for php-worker  Cannot start service php-worker: oci runtime error: container_linux.go:247: starting container pr
ocess caused "chdir to cwd (\"/etc/supervisor/conf.d/supervisord.conf\") set in config.json failed: not a directory"
: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the ex
pected type
```

Mahmoudz:
@cabrerabywaters @cjmaxik so what's the status here! shall we merge?!

cabrerabywaters:
I'm going to correct both mistakes and make the PR again. totally right the workdir is wrong and the volume should be applications.Hope to fix all my PRs tonight.





On Thu, Mar 16, 2017 at 2:34 PM -0300, "Mahmoud Zalt" <notifications@github.com> wrote:












@cabrerabywaters @cjmaxik so what's the status here! shall we merge?!

Mahmoudz:
@cabrerabywaters Thank you if you want to submit a new PR instead of committing to this, just mention closing this in your new one. Thanks in advanced.

cabrerabywaters:
@Mahmoudz  Ok! Fixed the container with @cjmaxik 's comments. 
Dockerfile now has a the proper Workdir and Volumes point to the applications volume. I just give it a try and works OK.

cjmaxik:
@cabrerabywaters there is a changed structure of docker-compose.yml with environment variables, could you resolve conflicts?

cabrerabywaters:
I just realized that...is this new? I just fetched uptream yesterday.
Sure..will create a new PR to fetch master again.
Regards

cjmaxik:
@cabrerabywaters yep, there is new. Actually, you don't need to create a new PR. Just click Resolve conflicts here and you're good to go.

cabrerabywaters:
@cjmaxik all set.
Also changed the links: statement to depends_on:

Mahmoudz:
@cabrerabywaters looks awesome, shall I merge!?

cjmaxik:
@Mahmoudz looks good for the merging. 👍

docker-compose.yml
```diff
@@ -70,7 +70,16 @@ services:
         extra_hosts:
             - "dockerhost:${DOCKER_HOST_IP}"
         environment:
-            - PHP_IDE_CONFIG=${PHP_IDE_CONFIG}
+            - PHP_IDE_CONFIG=serverName=laravel
```
cjmaxik: Wait a second here. PHP_IDE_CONFIG=${PHP_IDE_CONFIG} is the correct one.

cabrerabywaters:
done...sorry about that!..next PR is going to be better =)

docker-compose.yml
Mahmoudz: I havn't tried and never did before, but supposedly it should just replace that text from the env, I see no reason why shouldn't unless it never did! :P

Mvzundert:
@Mahmoudz I feel like having a worker run in PHP is very ineffective... PHP isn't meant to run infinitely (not exactly what we're doing here but still).  I'm currently working on a Ruby cli worker (for rabbitmq but it could listen to beanstalkd as well off course) and I know other stuff like: https://github.com/ricbra/rabbitmq-cli-consumer exists. I've run into a lot of issues running workers in PHP though so I'm curious as to what made you guys decide to build it in php.

cjmaxik:
@Mvzundert, this worker is Laravel-way solution directly from the Laravel developers. If you have more reliable option, go ahead 😃

Mvzundert:
@cjmaxik Thanks, I realized that it was from laravel itself... I'll see what I can cook up myself to support something like this, thanks for the reply!

baskan:
im having difficulty with supervisorctl commands the way we build php-worker from "php:<version>-alpine" cannot send commands like: 

> docker-compose run php-worker supervisorctl restart xyz

which is super convenient when you change files and need to restart workers itself. any suggestions or hacks on this?

cjmaxik:
@baskan, 
> php artisan queue:restart

baskan:
@cjmaxik thanks for the response, that will do for most people! 

but in my case; I **really** need to restart (I'm booting up a reactphp daemon with supervisor to be exact)

PR #675, opened 2017-03-15
Multiple projects improvement

Mvzundert:
I've Updated env-example and removed .env.example when using .env.example as well and example, it doesn't show up (it's a hidden file) when using non commandline tools. This could be confusing for people trying to set it up for the first time in my opinion.


Mahmoudz:
Can you send the same PR to branch [`multiple-projects-improvement `](https://github.com/laradock/laradock/tree/multiple-projects-improvement) instead of `master`. Thanks,

Mvzundert:
@Mahmoudz Whoops, sure thing, imcoming ;)

PR #676, opened 2017-03-15
make docker-compose available global | <laradock up -d nginx> from an…

31H0B1eV:
Hope I have followed correctly with contributing guide. This is just a little notice in docs about how to make available usage of `docker-compose <anything>` from anywhere. 

example: `laradock up --build -d nginx mysql `

### DIFF
--- a/docs/_settings/content/getting-started/index.md
+++ b/docs/_settings/content/getting-started/index.md
@@ -108,7 +108,13 @@ Do the same for each project `project2.conf`, `project3.conf`,...
 
 7 - Create your project Databases. Right now you have to do it manually by entering your DB container, until we automate it soon.
 
+8 - Sometimes you may want to call `docker-compose <anything>` from anywhere on your filesystem. You can do this on Mac / Linux systems by adding a Bash function to your Bash profile (just add this lines inside your ~/.bashrc).
 
+```bash
+function laradock() {
+    ( cd ~/laradock && docker-compose $* )
+}
+```
 
 
 
--- a/docs/contributing/index.html
+++ b/docs/contributing/index.html
@@ -10,7 +10,7 @@
     <meta name="viewport" content="width=device-width,user-scalable=no,initial-scale=1,maximum-scale=1">
     <meta http-equiv="X-UA-Compatible" content="IE=10" />
     <title>Contributing - Laradock</title>
-    <meta name="generator" content="Hugo 0.18.1" />
+    <meta name="generator" content="Hugo 0.16-DEV" />
 
     
     <meta name="description" content="Laradock documentations.">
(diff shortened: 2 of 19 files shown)


docs/_settings/content/getting-started/index.md
```diff
@@ -108,7 +108,13 @@ Do the same for each project `project2.conf`, `project3.conf`,...
 
 7 - Create your project Databases. Right now you have to do it manually by entering your DB container, until we automate it soon.
 
+8 - Sometimes you may want to call `docker-compose <anything>` from anywhere on your filesystem. You can do this on Mac / Linux systems by adding a Bash function to your Bash profile (just add this lines inside your ~/.bashrc).
 
+```bash
+function laradock() {
+    ( cd ~/laradock && docker-compose $* )
```
31H0B1eV: Not sure maybe better notice that `cd ~/laradock` must be a path of cloned repo.

mikeerickson:
The problem with this addition is when you have multiple docker projects (outside of laradock). Making this global could effectively up the wrong docker config.   I don't know that we want to make this a built-in command

31H0B1eV:
@mikeerickson Not sure how to reproduce this case. This command call docker-compose inside laradock folder and there is no matter how much docker projects you have in your system. Maybe I don't understand something?

mikeerickson:
@31H0B1eV What I am saying is, you could have `laradock` in any number of directories (not specific to `~/laradock` .   When executing `docker-composer ...` it will use the `docker-compose.yml` located in the directory where you execute `docker-compose` (ala the way `artisan` works).

Unlike Homestead (where you have most likely extracted the laradock method), laradock is not specifically a global tool.   

- homestead can work globally as it has a single `Homestead.yml` which configures all directories etc.

- laradock is not global by nature, thus should not be a globally access tool

31H0B1eV:
@mikeerickson And again did you try to reproduce it before report that it can raise a bug?  I don't know how to explain this and I think that will be better leave it as is. If you think it can cause a problem -ok. But cases which you have described is impossible at all :)

example:

1) Any project with laradock inside:
 - `docker-compose up -d nginx`
 - `laradock up -d nginx` here will do nothing for this project and this is just mean `cd ~/laradock && docker-compose up -d nginx` 
  
2) project wich configured for use global laradock:
 -  `laradock up -d nginx` from any place and there is no matter do you have docker-compose.yml or not.
 - if you call here  `docker-compose up -d nginx` -you do something wrong.

P.S. I think you don't understand simple thing, at first homestead also can be used not only like global stuff, but also for each project separately. If you call `docker-compose <anything>` yes it will call it based on docker-compose.yml in same folder, but if you use command that i have propose you never call `docker-compose <anything>` for projects (only if you need it and set up another laradock instance, here you will use it as always), all that you do is call `laradock <anything>`. This is the same if you call two commands:
1 `cd ~/laradock`
2 `docker-compose <anything>`

Please before report a problem try to reproduce this problem and give a little description for steps which must be done.

mikeerickson:
@31H0B1eV Well, I stand (well I am sitting) corrected it appears.   ;-)

I am going to give this PR a try and see how it works.  My apologies for making some seemingly bad assumptions.

PR #688, opened 2017-03-17
re-added .gitignore for nginx dir, still do not want configs in git

Mvzundert:
It seems weird to now include the conf files in git... I couldn't really find where this got removed, but I feel it still needs to be there.


a-h-abid:
Why was .gitignore removed?

Mvzundert:
@a-h-abid  I have no idea, but I'm requesting it be re-added.

PR #700, opened 2017-03-20
fix some errors in mysql setup file for latest mysql image

picospuch:
I found the replacement of MYSQL_DATABASE for 'default' in mysql/startup file will cause sql syntax error. No default database will be created on the creation of the docker container.

the mysql version pulled by docker is 5.7.17-1debian8.

and, according to https://dev.mysql.com/doc/refman/5.7/en/grant.html
image

It's likely better to append IDENTIFIED BY to CREATE USER statement.


alchimik:
see #649, file mysql/startup will be removed

PR #712, opened 2017-03-24
adding MsSQL DATABASE support

cabrerabywaters:
@Mahmoudz  Here is the container for SQL Server requested in [670](https://github.com/laradock/laradock/pull/670) I just tested and its working OK...please give it a try and let me know what you think.


docker-compose.yml
```diff
@@ -156,6 +156,20 @@ services:
...
+
+    mssql:
+      build:
+        context: ./mssql
+      environment:
+        - MSSQL_DATABASE=${MSSQL_DATABASE}
+        - SA_PASSWORD=${MSSQL_PASSWORD}
+        - ACCEPT_EULA=Y
+      volumes:
+        - mssql:/var/lib/mysql
```
thorerik: MSSQL on linux uses /var/opt/mssql, not /var/lib/mysql

cabrerabywaters: @thorerik You are right thank you!

Issue #736, opened 2017-04-03
Can't connect to mariadb


vitoo:
Hello,

I can't make mariadb work with laradock v5.0.3, with stock docker-compose.yml  and  .env

It works with mysql throug adminer (host: mysql, port: 3306, root/root)  and inside the mysql container.

But i can't connect to mariadb from adminer : (host: mariadb, port: 3306, root/root) 
`SQLSTATE[HY000] [1045] Access denied for user 'root'@'172.18.0.4' (using password: YES)`

And even inside the mariadb container : 
```
$ docker-compose exec mariadb bash
root@88cf42a7d05e:/# mysql -uroot -proot
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)
```
MariaDB container log : 

```
mariadb_1             | 2017-04-03 19:25:16 140329248704448 [Note] Server socket created on IP: '::'.
mariadb_1             | 2017-04-03 19:25:16 140329248704448 [Warning] 'proxies_priv' entry '@% root@077c23180bc8' ignored in --skip-name-resolve mode.
mariadb_1             | 2017-04-03 19:25:16 140329248704448 [Note] mysqld: ready for connections.
```

Any idea to fix this ?

Thanks for your help!

winfried-van-loon:
Within the dotenv file, you can still use the `MARIADB_USER` user and password. I'm having the same issue with the root user. There seems to be a problem in the config file: https://github.com/laradock/laradock/blob/master/docker-compose.yml#L207

Mahmoudz:
Closed #737 by @winfried-van-loon. Thanks,

winfried-van-loon:
@Mahmoudz I'm afraid it was only part of the solution, with the fixes included I wasn't able to actually connect. Could you give it a try?

vitoo:
Maybe the issue is due to the volume ?
The same volume are mounted on the host machine ? 

`/var/lib/mysql
`

winfried-van-loon:
Well, the "default" user does work, so I doubt it's in the network or connectivity of the volumes. I've checked the `docker-compose.yml` for related issues, haven't found any. Good idea, though!

winfried-van-loon:
May be related to https://github.com/docker-library/mariadb/issues/21 and a possible temp solution could be to reset the root user password until we find the real problem&solution: https://github.com/docker-library/mariadb/issues/99

vitoo:
Yes we need to reset the mariadb volume for sure

[...]

PR #737, opened 2017-04-03
Fixing MariaDB root password variable. #736

winfried-van-loon:
Should fix the issue with the MariaDB root password.


winfried-van-loon:
This doesn't seem to be the only problem. Using this fix, still not able to connect.

PR #744, opened 2017-04-05
Fixing specified Dockerfile name

gentcys:
`ERROR: Cannot locate specified Dockerfile: dockerfile-71`


gentcys:
@Mahmoudz Please revert this request, there is another place to fix.

Issue #747, opened 2017-04-05
Error in php-fpm ZIP ext


winfried-van-loon:
My environment:

- MacOS 10.12.4
- Docker for Mac Community 17.03.1-ce-mac5 (16048)
- Using [d4m-nfs](https://github.com/IFSight/d4m-nfs)

To reproduce the issue:

- Set `PHP_VERSION=56`
- Set workspace Dockerfile to 70 to avoid those errors: `dockerfile: "Dockerfile-70"`
- Using the following command:

```
docker-compose up -d --force-recreate --build apache2 mariadb
```

Related output will be the following, containing the error.

```
Step 11/36 : RUN if [ ${INSTALL_ZIP_ARCHIVE} = true ]; then     pecl install zip &&     docker-php-ext-enable zip ;fi
 ---> Running in bcf5e48386a2
downloading zip-1.14.0.tgz ...
Starting to download zip-1.14.0.tgz (248,476 bytes)
.............................done: 248,476 bytes
8 source files, building
running: phpize
Configuring for:
PHP Api Version:         20131106
Zend Module Api No:      20131226
Zend Extension Api No:   220131226
building in /tmp/pear/temp/pear-build-defaultuserXxuUdq/zip-1.14.0
running: /tmp/pear/temp/zip/configure --with-php-config=/usr/local/bin/php-config
checking for grep that handles long lines and -e... /bin/grep
checking for egrep... /bin/grep -E
checking for a sed that does not truncate output... /bin/sed
checking for cc... cc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether cc accepts -g... yes
checking for cc option to accept ISO C89... none needed
checking how to run the C preprocessor... cc -E
checking for icc... no
checking for suncc... no
checking whether cc understands -c and -o together... yes
checking for system library directory... lib
checking if compiler supports -R... no
checking if compiler supports -Wl,-rpath,... yes
checking build system type... x86_64-unknown-linux-gnu
checking host system type... x86_64-unknown-linux-gnu
checking target system type... x86_64-unknown-linux-gnu
checking for PHP prefix... /usr/local
[...]

PR #751, opened 2017-04-05
Fix PHP Zip library installation

david-perez:


winfried-van-loon:
Cool, this looks like it may just fix #747 👍

PR #803, opened 2017-04-12
Added support for jenkins official container

activeobjects:


Mahmoudz:
That's a lovely PR, thanks. 

i think you missed re-generating the documentation site, after modifying the Readme files! 
Note: this won't happen automatically yet. I hope someone has some time to integrate a CI tool to automate that process but meanwhile it needs to be done manually since we're using Hugo. The steps of doing that are in the contributing guide.

If you don't want to regenerate it, just wait until someone modifies it again and do that, then you're updates will appear in the site. 

Best,

mikeerickson:
@Mahmoudz @activeobjects I will be updating document (need to document setup and configuration guide for Apache2 -- seeing as @Mahmoudz doesnt like Apache2)   I will make sure it includes the updated guide for this PR as well

PR #810, opened 2017-04-16
set data save path(mysql/redis/memcached..)

wxb:
save data file to local


a-h-abid:
Can I know what was the reason for this PR? What benefit it brings?

wxb:
save database file，e.g:  persisted mysql/redis  data to hard disk

PR #837, opened 2017-04-21
Upgrade hugo to 0.20.2

dtunes:
This fixes issues I had recently when using hugo 0.20* to generate the docs, where index.md would create a path like /docs/introduction/index/index.html, adding uglyurls=true fixes this.

I also fixed the path to photos in the guide, even on the site all photos* were Not Found (404), fixed all paths to images/photos*


dtunes:
Ok, need to review this, if you check the guides, the images still not showing. Locally, the path images/photos worked. However, when hosted on gh-pages, its not working. Removing the guide/ from the path seems to fix it, might have to look into the hugo's documentation.

Mahmoudz:
@dtunes yep I think that guide is the problem, I also see different results between my local host and the online doc. And I'm using the same version!

PR #848, opened 2017-04-23
add support for php-fpm 5.3

shemayas:
add a option to use php 5.3
Iv'e used helder/php-5.3 image as there is no official php-fpm 5.3 image.

### DIFF
--- a/env-example
+++ b/env-example
@@ -14,7 +14,7 @@ DATA_SAVE_PATH=/tmp
 
 ### PHP version
 # Applies to the Workspace and PHP-FPM containers (Does not apply to HHVM)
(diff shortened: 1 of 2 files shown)


thorerik:
PHP 5.3 is EOL, so IMO that is far outside what should be supported. You are free to use it, but I don't think it's something we should maintain and support in any way.

Mahmoudz:
@shemayas thanks for submitting this PR. I do agree with @thorerik and I suggest you include that in your own fork of this repo, as it might not be useful for many other users, especially that maintaining a PHP version is not so smooth task.. That's why we dropped older php version support.

Issue #849, opened 2017-04-23
Apache2 Docroot


kevin-coyle:
Just done a recent clone of the project and I've noticed that the document root is set to /var/www instead of /var/www/public in the apache2 container.

Is there a variable in the .env file to control this? I can see
```
ENV WEB_DOCUMENT_ROOT=/var/www
```
In the Dockerfile, it used to be /var/www/public

winfried-van-loon:
This should just work. For the NGINX container, it looks for the public directory, for the Apache container, it does not. It should either be both looking for the public directory, or both not looking for the public directory.

kevin-coyle:
Hmm they seem to be different now

winfried-van-loon:
This should have been fixed in the #850 pull request. If you're still experiencing this issue, let us know. 👍

vladthecodeimpaler:
Hi Guys,

apache2 doesnt work by default again, whats the reason for the revert?
```
DocumentRoot /var/www/public
is now
DocumentRoot /var/www/
```

Mahmoudz:
@vladthecodeimpaler cause without creating the `public` folder the container causes errors. Without this it will be working same like NGINX, and reading all the projects on the /var/www/ which is the directory where the Laradock folder exist.

benkelukas:
Hi, I have the same problem, will this be configurable in .env in the future? Right now i have rewritten it in default.apache.conf, but that does not seem right :)

Thanks

PR #850, opened 2017-04-23
Updating webroots to have the same defaults

winfried-van-loon:
Caddy, NGINX and Apache2 should have the same default application paths. The `public/` path suffix is now added in the dotenv file to be able to use Laradock for non-laravel projects as well. Also, updated the documentation.

This should fix issue #849.

Update: I can't spell today. Apologies for changing the title two times. 🔢


mikeerickson:
@winfried-van-loon Can you please address the (new) merge conflicts?  I will get this merged as soon as it is clean :-)

Issue #792, opened 2017-04-11
Workspace V8JS is broken


winfried-van-loon:
When enabling the following option, a whole lot of errors appear.

```
WORKSPACE_INSTALL_INSTALL_V8JS_EXTENSION
```

Honestly, I'm not quite sure how V8JS works and how to fix the issue. 

The log/errors that happened are pasted here, as it's very long I haven't posted it within the issue description: 
https://paste.laravel.io/yPY0j

**Update:** This args may have been renamed to: ```WORKSPACE_INSTALL_V8JS_EXTENSION```

PR #852, opened 2017-04-23
Fixing typo in the workspace V8JS args

winfried-van-loon:
Fixing typo in the workspace V8JS args. 

Please note that while this is related to #792, it does **not** fix that issue.


winfried-van-loon:
Clicking on buttons is hard, today. This pull request was in the wrong repo. New pull request with the same changes can be found here: #854

PR #888, opened 2017-05-04
Update Redis Docker File

tristanbailey:
The dir for the conf seems to be missing from the base image, so the copy fails

Also if the server does not start with the conf as parameter it will not pick it up. Which leaves Redis wide open to attack by default:

Example try to telnet to your ip on the redis port for this container:   
telnet 192.168.1.11 6379  
echo "Hey no AUTH required!"

### DIFF
--- a/redis/Dockerfile
+++ b/redis/Dockerfile
@@ -2,10 +2,13 @@ FROM redis:latest
 
 MAINTAINER Mahmoud Zalt <mahmoud@zalt.me>
 
+## For security settings uncomment, make the dir, copy conf, and also start with the conf, to use it
+#RUN mkdir -p /usr/local/etc/redis
 #COPY redis.conf /usr/local/etc/redis/redis.conf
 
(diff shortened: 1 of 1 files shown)


tristanbailey:
For people who see this later, also do this yourself, as people or bots can telnet in and install ssh keys and the like from redis in very few lines of normal commands.   
Explainer post:   
http://antirez.com/news/96

Copy in the latest conf   
http://download.redis.io/redis-stable/redis.conf

Edit the "requirepass" and "bind" lines, the pass needs to be complex, as redis can try so many passwords a second its ridiculous. The bind I found worked on a Digital Ocean server was to 0.0.0.0 rather than 127.0.0.1 or the main IP.   

bind 0.0.0.0   
requirepass something-very-long-and-complex-pass123@

winfried-van-loon:
Should this security measure not be enabled by default?

Issue #855, opened 2017-04-23
Enabling workspace/aerospike causes fatal error 


winfried-van-loon:
Enabling the Aerospike arg for Workspace causes a fatal error for PHP. This is tested for the PHP7.1 version, which is now the default.

Running `php -v` within the Workspace bash returns:
```
PHP Fatal error:  Cannot redefine class constant Aerospike:: in Unknown on line 0
```

winfried-van-loon:
May not be a Laradock issue, there is an open issue about this in the Aerospike repo:
https://github.com/aerospike/aerospike-client-php/issues/118

bestlong:
PHP_VERSION=71
If PHP_FPM_INSTALL_AEROSPIKE_EXTENSION and PHP_FPM_INSTALL_OPCACHE then
build failed

winfried-van-loon:
Should be fixed now by #906. However if you're still experiencing the issue, feel free to leave a comment.

JonayMedina:
Hi, i having error running docker-compose build --no-cache 

> ERROR [laradock-aerospike internal] load metadata for docker.io/library/aerospike:latest

> [laradock-aerospike internal] load metadata for docker.io/library/aerospike:latest:
> ------
> failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to create LLB definition: docker.io/library/aerospike:latest: not found

PR #897, opened 2017-05-05
Upgrede aerospike-client-php  version to 3.4.14

bestlong:


winfried-van-loon:
Sadly, the Aerospike issue #855 isn't fixed by this pull request. Nevertheless, it's an excellent idea to upgrade to the latest version available.

```
Installing shared extensions:     /usr/local/lib/php/extensions/no-debug-non-zts-20160303/
[91mPHP Fatal error:  Cannot redefine class constant Aerospike:: in Unknown on line 0
[0m[91merror: /usr/src/php/ext/opcache/opcache.so does not exist
```
(from the Travis CI logs of 71 container)

PR #901, opened 2017-05-06
Split Travis-CI build jobs and fix php-fpm mssql

bestlong:
One job build for one PHP version, to wait a long time.
so, split build job for each service. run fast and short log.

php-fpm MSSQL use stable version and fix error path.


bestlong:
Finally only left php-fpm on PHP71 build failed

PR #902, opened 2017-05-06
Nginx Proxy image

dlnsk:
Nginx Proxy is magnificent image!
No more juggling with ports! Start nginx, apache as many as you like at the same time with different PHP versions (if you want). Phpmyadmin, adminer, node may work simultaneously on beautiful domains instead of horrible ports.

It has more options. See: https://github.com/jwilder/nginx-proxy

### DIFF
--- a/DOCUMENTATION/content/documentation/index.md
+++ b/DOCUMENTATION/content/documentation/index.md
@@ -708,10 +708,44 @@ More details about this [here](https://github.com/jenssegers/laravel-mongodb#ins
 
 
 <br>
+<a name="Use-Nginx-Proxy"></a>
+## Use NGINX Proxy
+
+It starts automatically when you start nginx, apache2, phpmyadmin, adminer or pgadmin containers. So don't worry.
+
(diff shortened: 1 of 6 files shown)


Bonobomagno:
Omg! I was searching this for a while. Thank for the commit!

Mahmoudz:
@dlnsk that's looks very nice could you please update the docs "include the container in the list of available containers, for people to know it exist..", also "you can add a little section explaining for other users how to use or in case there's anything special they need to know.." Thanks

dlnsk:
@Bonobomagno 
> need some fix becouse of 503 error
It works fine. Be sure that you were update your own .env file

haringsrob:
I have updated my master to have the merge conflicts resolved: https://github.com/haringsrob/laradock

Bonobomagno:
Why still pending?

Mahmoudz:
@Bonobomagno this is great feature but this PR has some conflict, and needs serious testing by someone to confirm it's working. Can you help with this, basically just fix the conflict and test it locally, make sure it's not affecting other things. Then let us know to merge it quickly. Best,

gulien:
@Mahmoudz @dlnsk Jwilder provide a awesome reverse-proxy for sure! That's said, I've you considered using https://traefik.io/ instead?

makowskid:
Massively outdated PR, untested, a lot of conflicts. Also might be not needed anymore, we already have Traefic image incorporated.

PR #903, opened 2017-05-07
Less log info

bestlong:
apt-get update info can ignore.


winfried-van-loon:
This is a simple, yet great addition. Thank you!

PR #920, opened 2017-05-08
COPY Caddyfile to docker image.

Lednerb:
The given Caddyfile would not be used if it's not copied to the image.


Lednerb:
Oh, I was wrong.
Of course it's handled by the `docker-compose.yml`.

PR #938, opened 2017-05-13
Percona service use DATA_SAVE_PATH env var

bestlong:
Same way mysql, mariadb.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -214,7 +214,7 @@ services:
         - MYSQL_PASSWORD=${PERCONA_PASSWORD}
         - MYSQL_ROOT_PASSWORD=${PERCONA_ROOT_PASSWORD}
       volumes:
-        - percona:/var/lib/mysql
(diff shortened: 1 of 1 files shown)


dtunes:
Don't like this approach, hence why I kept it as a data volume, for container data like mysql, redis, elastic etc.. I'd use data volumes instead of mounting the a host dir on the containers.

springjk:
@dtunes i have the same confusion like you say. 

Why i create a lot of volumes, but no one inuse?

All container use the `DATA_SAVE_PATH`, when i do some change in `docker-compose.yml`, i must clear `DATA_SAVE_PATH`, i can't use `docker-compose down -v`, so, who is using the volume mysql? I hope I didn't miss anything for this config. is that a bug?

dtunes:
@springjk you're not missing anything, a long time ago people started using DATA_SAVE_PATH over volumes, I don't know I prefer volumes instead of mapping it to a local path, I think its cleaner.

PR #964, opened 2017-05-24
Dockerfile-56 fill missing MSSQL section

MyController:


bestlong:
Hi @MyController suggest ARG INSTALL_MSSQL=false

MyController:
done

Issue #1009, opened 2017-06-09
Mysql container fails on start


invasioncredible:
Latest docker - latest laradock (master)

If i do `docker-compose up -d nginx php-fpm mysql phpyadmin `
nginx - almost default config
php - 7.0
Mysql - simple configuration
Phpmyadmin - no custom config

it starts, and then the mysql container falls, and i have to do
`docker-compose up -d mysql `
and only then it properly runs. 
I didn't have this problem with the previos versions of laradock. 

Not a fatal bug, but i would be happy if you fix this)

dtkoval:
Just try to use [docker-machine-nfs](https://github.com/adlogix/docker-machine-nfs) if you use mac

Mahmoudz:
@invasioncredible sometimes it does that in the first attempt! try `docker-compose down` then run your containers again. And let us know if it works.

nguyencongkrb:
I have fixed it. Disable AIO

https://stackoverflow.com/questions/48239668/fails-to-initialize-mysql-database-on-windows-10

PR #1010, opened 2017-06-09
fix(issue #1009): fixed the issue

invasioncredible:
I fixed a little issue. Not sure, if it's a good idea to make it this way, but it worked for me)


Mahmoudz:
@invasioncredible problem has been solved. Thank you.

Issue #1004, opened 2017-06-08
ElasticSearch for Docker discontinued, use official repository


ItaloBC:
### Info:
- Docker version: [all] 
- Laradock commit: d0c0a8895efcc36d89f03e8869ad971bd2704d4a
- System info (Mac, PC, Linux): Unrelevant
- System info disto/version: Unrelevant

##### What seems to be going wrong?

ElasticSearch for Docker has been discontinued in the Docker Official Registry.

Next releases will be maintained under on `docker.elastic.co` exclusively.

[Instructions how to set up the new maintained image are in elastic.co](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html).

Also, current releases creates a Docker image for ElasticSearch but it halts immediately. This solution also fixes that particular problem.
_____

### Relevant Code:

To fix the problem this should be done

./elasticsearch/Dockerfile
```
FROM docker.elastic.co/elasticsearch/elasticsearch:5.4.1

EXPOSE 9200 9300
```
** note, the version should be the latest stable. Currently is 5.4.1. using `latest` instead the version number won't work to pull the latest image version.

./docker-compose.yml
```
elasticsearch:
      build: ./elasticsearch
      volumes:
        - elasticsearch-data:/usr/share/elasticsearch/data
        - elasticsearch-plugins:/usr/share/elasticsearch/plugins
      environment:
        - cluster.name=laradock-cluster
        - bootstrap.memory_lock=true
        - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
      ulimits:
        memlock:
          soft: -1
          hard: -1
      mem_limit: 1g
      ports:
        - "${ELASTICSEARCH_HOST_HTTP_PORT}:9200"
        - "${ELASTICSEARCH_HOST_TRANSPORT_PORT}:9300"
      depends_on:
        - php-fpm
      networks:
        - frontend
        - backend
```

You *could* add some parameters under the `.env`
- Xms and Xmx, determine the JVM Heap size for garbage collection, basically.
- mem_limit: basically, how much memory will use elasticsearch as a whole. 1GB sounds normal, but for devolping it could be pushed to `512m`.

winfried-van-loon:
Thank you for looking up that much information about the subject! Would you be kind enough to submit a pull request? That way, the credits go to you, as well. 🥇

ItaloBC:
I'll do.

I'm testing v5.5.0 so I will commit over that.

ItaloBC:
[...]

PR #1013, opened 2017-06-12
Kibana - Change to official repository

ItaloBC:
Elastic.co announced they will pull out their official images from the Docker Registry and user their own. There will be no updates starting 6/20/2017.

Pulling needs explicit version. Currently 5.4.1


winfried-van-loon:
Related issue: #1004

PR #1015, opened 2017-06-12
Set Timezone

ardziej:
Properly set timezone for mysql container


ardziej:
@winfried-van-loon what about rest of containers ? cause it's can be a problem for all of the containers. eg. queues. mssql, php-fpm. Is it the only one way to set timezone for all of them, one by one ? I can make a one commit for it, if it is the best idea to fix timezone problem in laradock.

Best! 👍

PR #1050, opened 2017-07-07
Allow specifying the version of MySQL to install in .env file.

hackel:
The recent upgrade to MySQL 8.0 (development release!) prompted me to add this change and I thought it would probably be useful to everyone.

### DIFF
--- a/DOCUMENTATION/content/documentation/index.md
+++ b/DOCUMENTATION/content/documentation/index.md
@@ -158,7 +158,7 @@ You might use the `--no-cache` option if you want full rebuilding (`docker-compo
 
 ## Docker-Sync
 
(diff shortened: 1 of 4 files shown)


Mahmoudz:
@hackel have you tested this?? I don't think this ever works!!

hackel:
@Mahmoudz Yes indeed, it's working fine for me with MySQL 5.7 installed on my container right now.  I struggled with this for a while myself before I finally figured it out.  The key was using build_args instead of environment variables.
See: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

Mahmoudz:
@hackel I've just tested it and it's working as you said, that's really cool feature, wasn't possible in the past. It will help us define the php version and remove the multiple docker files. Would you be interested in doing that with some if-else in the single docker file to install the correct plugins based on the defined version?

PR #1058, opened 2017-07-16
Integrating docker sync

arianacosta:
I think Laradock is a fantastic project, unfortunately, after setting it up I realized it that was performing very poorly on OS X. In my benchmarks, it is slower than MAMP and even Vagrant. Loading a page would take ~1.2 seconds, which makes laradock less than ideal for development on a Mac.

Looking for solutions, I came across [docker-sync](https://github.com/EugenMayer/docker-sync) which processed the same page in about ~180ms (6X faster than vanilla laradock). However, setting it up was very cumbersome and took a lot of time, because of the lack of good documentation, and several problems with user permissions and syncing methods. For that reason, I have decided to integrate it with Laradock and give back to this awesome community. 

In short, these are some of the features of my contribution:

- Provides an easy out-of-the-box solution to setup docker-sync.
- Does not affect previous laradock setups.
- Enables laradock to run with or without docker-sync on demand.
- Incorporates an optional `sync.sh` to easily install, start and stop laradock with docker-sync.
- Maintains configurations dry.

All the details are in the documentation, but the usage is simple, just run:

```
./sync.sh install
./sync.sh up nginx mysql
./sync.sh down
```
These are the available options:
```
   install		 Installs docker-sync gem on the host machine.
   up [services]	 Starts docker-sync and runs docker compose.
   down			 Stops containers and docker-sync.
   bash			 Opens bash on the workspace.
   sync			 Manually triggers the synchronization of files.
   clean		 Removes all files from docker-sync.
```
##### Make sure you completed the basic 3 steps below:

- [x] I've read the simple [Contribution Guide](http://laradock.io/contributing).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /data
 .env
 /.project
+.docker-sync
--- a/DOCUMENTATION/content/documentation/index.md
+++ b/DOCUMENTATION/content/documentation/index.md
@@ -156,21 +156,82 @@ You might use the `--no-cache` option if you want full rebuilding (`docker-compo
 <br>
 <a name="Docker-Sync"></a>
 
-## Docker-Sync
+## Speed up with docker-sync
 
-Docker on the Mac [is slow](https://github.com/docker/for-mac/issues/77), at the time of writing. Especially for larger projects, this can be a problem. The problem is [older than March 2016](https://forums.docker.com/t/file-access-in-mounted-volumes-extremely-slow-cpu-bound/8076) - as it's a such a long-running issue, we're including it in the docs here. 
+Docker for Mac is [slow](https://github.com/docker/for-mac/issues/77) due to poor performance when the application accesses files shared with the host machine. 
+One solution is to use [docker-sync](https://github.com/EugenMayer/docker-sync). 
 
-The problem originates in bind-mount performance on MacOS. Docker for Mac uses osxfs by default. This is not without reason, it has [a lot of advantages](https://docs.docker.com/docker-for-mac/osxfs/).
+In simple terms, docker-sync creates a docker container with a copy of all the application files that can be accessed very quickly from the other containers. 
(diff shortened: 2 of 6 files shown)


Mahmoudz:
@arianacosta looks pretty cool. Thanks for the contribution. 

I think the docs aren't accurate I think it should be `./sync.sh docker-sync-stack up` instead `docker-sync-stack start`!

arianacosta:
Hi! Thanks for the quick reply. I tried using `docker-sync-stack up`, but it doesn't allow you to specify the services as arguments as the usual `docker-compose up <services>`. In my tests, it would just start all services. So, I wanted to keep it as flexible and as close as possible to the regular `docker-compose` command which allows people to run only the containers they need at that moment. What do you think it should be?

Mahmoudz:
@arianacosta totally agree on keeping the commands similar to docker. My concern was with the missing `./sync.sh ` and when I tried the up command it didn't work! Maybe I did something wrong. Could you please add to the docs, steps like this:

1. run command xxx
2. do this
3. do that..

With some details on what's going on behind the scene.

Thanks,

arianacosta:
@Mahmoudz No problem, I will improve the documentation even more, and add the info of the commands without the `.sync.sh` helper. Also, if you could provide the error that you are seeing, that would be great.

Thank you

realshadow:
@arianacosta This might not be good place to ask this. But I don't see any performance improvment whatsoever. Tried all strategies, request time fluctuates from 4-6 seconds - same as vanilla laradock. Any idea what could be the cause? I can get down to 500ms with dinghy

hmaurer:
Unless I am mistaken docker-compose should be started using a flag (or env var) to load `docker-compose.sync.yml` instead of `docker-compose.dev.yml`? Using `./sync.sh up` did not work for me. After some googling I tried `docker-compose -f docker-compose.yml -f docker-compose.sync.yml up`, which seems to work perfectly. Am I missing something?

PR #1117, opened 2017-08-31
Fix Php Redis extension installation

masterjus:

Currently php-redis extension doesn't installs to php 71 workspace 

The problem is because during installation it asks `enable igbinary serializer support?` and doesn't install extension because didn't get response.


Mahmoudz:
@masterjus can you add this to the other PHP versions?

masterjus:
ok

PR #1133, opened 2017-09-13
apache2 default settings

tpalts:
This pull request includes updates to the apache2 config meant ensure that the package works as a new user would expect. Right now new user has to edit the apache2 conf for a single-site-setup to work with apache2. This update fixes that.

### DIFF
--- a/apache2/sites/default.apache.conf
+++ b/apache2/sites/default.apache.conf
@@ -1,5 +1,5 @@
 <VirtualHost *:80>
-  ServerName laradock.dev
+  ServerName localhost
   DocumentRoot /var/www/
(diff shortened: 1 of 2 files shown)


alchimik:
reason of removing the public -
 https://github.com/laradock/laradock/commit/2af40b0ba107a86b57710226fb3b2c29b309f164#commitcomment-23570558

Mahmoudz:
@tpalts it's not a problem we needed to make the default setup compatible with multiple projects. Cause more people uses Laradock as a central development tool for all their projects. And in all cases if you want to use it for single project you'll have to edit some files such as removing unused containers.. so editing the server config file is very normal.

Issue #621, opened 2017-02-21
Service 'nginx' failed to build


infacq:
I just about to install Laradock but it failed when hit the abovemention error

```bash
ERROR: Service 'nginx' failed to build: The command '/bin/sh -c apk update     && apk upgrade     && apk add --no-cache bash     && adduser -D -H -u 1000 -s /bin/bash www-data     && rm /etc/nginx/conf.d/default.conf     && echo "upstream php-upstream { server ${PHP_UPSTREAM}:9000; }" > /etc/nginx/conf.d/upstream.conf' returned a non-zero code: 1
```

where do I go from here

sonialiang:
I am new to docker.
I met with same problem on My AWS in us-east-1 region.
I spent 2 days to solved the issue.

The reason is because the repository mirror site.
People also discussed about this. Refer #609  

Try delete the merged line. And run it again.

fedemotta:
@sonialiang It seems that this mirror site works fine from china, maybe we can create a ENV variable to select the proper site to download the alpine links and let the user decide which one to use

PR #1251, opened 2017-11-14
Set environment to change source.

mouyong:


mouyong:
Such as the [issue](https://github.com/laradock/laradock/issues/621) [@fedemotta](https://github.com/fedemotta)  said: 

> It seems that this mirror site works fine from china, maybe we can create a ENV variable to select the proper site to download the alpine links and let the user decide which one to use.

So with this pull request

PR #1265, opened 2017-11-21
Specifying included docker-compose files in the .env is more flexible

jeff-h:
Specifying included docker-compose files in the .env is more flexible than hardcoding them in the shell script.

### DIFF
--- a/env-example
+++ b/env-example
@@ -7,6 +7,9 @@
 
 COMPOSE_FILE=docker-compose.yml:docker-compose.dev.yml
 
+# Replace the above with the following line if you wish to use docker-sync.
+# COMPOSE_FILE=docker-compose.yml:docker-compose.dev.yml:docker-compose.sync.yml
+
 ### Application Path ###################################################################################################
 # Point to your code, will be available at `/var/www`.
 
--- a/sync.sh
+++ b/sync.sh
@@ -59,7 +59,7 @@ if [ "$1" == "up" ] ; then
 
     print_style "Initializing Docker Compose\n" "info"
     shift # removing first argument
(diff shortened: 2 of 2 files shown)


jeff-h:
@stevenobird So sorry for the oversight. This kind of thing really annoys me in Docker... so many different fiddly things to change all the time! I **really** dislike the fact my PR means you cannot just use a copy of the supplied `env-example` on Windows and expect it to work :(

Since (I'm guessing) you're on Windows, would you be able to test `COMPOSE_PATH_SEPARATOR` and see if it can be set to `:` on Windows, or does that break everything?

`COMPOSE_PATH_SEPARATOR=:`

I assume it breaks things on Windows or it would come out-of-the-box with the same separator across all three platforms?

Relevant docs: https://docs.docker.com/compose/reference/envvars/#compose_file

@Mahmoudz Would you consider a PR changing the compose path separator to `#` (or something like it)?

stojankukrika:
Yes, It's work in windows. I just add `COMPOSE_PATH_SEPARATOR=: ` into .env file and it's start docker. Before that have that error `ERROR: .IOError: [Errno 22] invalid mode ('r') or filename: u'.\\docker-compose.yml:docker-compose.dev.yml'`

jeff-h:
Wow! That's both great news and also incredibly annoying. Thanks for testing it.

So... why is `:` not standard "out-of-the-box" in Windows docker-compose then?

The fix now seems to be simply to add `COMPOSE_PATH_SEPARATOR=:` into the default `env-example` file, would you agree? We can then remove any help notes regarding this from the documentation — it should just work on all three platforms without modification.

Issue #1268, opened 2017-11-23
ERROR: .IOError: [Errno 2] No such file or directory


xiaolony:
> docker-compose up -d nginx mysql phpmyadmin

get:

> ERROR: .IOError: [Errno 2] No such file or directory: u'.\\docker-compose.yml:docker-compose.dev.yml'

why?thanks

uighurbabbage:
Do you order that you execute it under the laradoc project?

codebykyle:
I also get this error, following the documentation on the documentation website. I am running Windows 10 with docker on hyper-v, on an existing laravel project. I added the git sub-module, and tried to run the command from documentation:

![image](https://user-images.githubusercontent.com/26160129/33169844-b5f08786-d078-11e7-8f45-71d969f842f6.png)

uighurbabbage:
It won't be windows10 environmental problems。
I didn't have this problem at Mac, centos7.4, and ubuntu16.04。
Are you sure that there are two files docker-compose.yml and docker-compose-dev.yml in the directory that executes this command in person, and if this is available, see if the docker version is low, and if this is not, it's likely to be a windows10 environmental problem

codebykyle:
The file is there
![image](https://user-images.githubusercontent.com/26160129/33170121-b32d6752-d079-11e7-896f-4c742685a688.png)

The docker version is 17.09.0-ce-win33 stable release, which I believe is the current stable release provided by docker. 
![image](https://user-images.githubusercontent.com/26160129/33170138-c9be772c-d079-11e7-978f-8d68a5c27153.png)

I have no issues running other docker containers. I do not believe this to be directly a windows 10 environmental problem.

codebykyle:
I also tried this in powershell just to be sure. I receive the same error, while the file is discoverable in the directory by hitting tab 
![image](https://user-images.githubusercontent.com/26160129/33170352-a27d6a5a-d07a-11e7-91a9-df3fdccd1b0d.png)

uighurbabbage:
The same version, I don't have this problem, that estimate is windows10 environmental problems, you try to use virtual machine installed linux environment to play。

codebykyle:
Sorry, I'm really looking to use this on windows, since my development environment requires use of specific windows programs. Docker should be able to run on Windows alright. Additionally, the documentation discusses use on Windows, which leads me to believe Laradock should be able to run just as well regardless of the platform.

There is something specific going on here. I will look into this and report back if I find any solution to the problem.

[...]

PR #1269, opened 2017-11-23
Clarified how to split multiple docker-compose files on Windows

francoisfaubert:
As seen in #1268, is it unclear in the latest version that COMPOSE_FILE must separate files with `;` on Windows when setting up Laradock.

This PR adds both a new comment in `.env-example` and on the Getting Started page.

Fell free to reject this PR if you'd rather automate the separator toggle instead of adding the details to the docs. Seems like a quick win at the moment though.

### DIFF
--- a/DOCUMENTATION/content/getting-started/index.md
+++ b/DOCUMENTATION/content/getting-started/index.md
@@ -162,6 +162,7 @@ cp env-example .env
 
 You can edit the `.env` file to chose which software's you want to be installed in your environment. You can always refer to the `docker-compose.yml` file to see how those variables are been used.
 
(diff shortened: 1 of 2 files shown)


jeff-h:
This is perfect, thanks @francoisfaubert 

@Mahmoudz — I am hoping you can roll this in asap, because my previous PR is causing lots of hassles for Windows users (I didn't realise Docker defaults to a different file delimiter on Windows than on Linux and macOS). This PR also fixes https://github.com/laradock/laradock/pull/1265

Harti:
Adding a `env-example.windows` file identical to the `env-example` file (except for the `:`) and adding "Note: Windows users are required to run `cp env-example.windows .env` instead" could be a more intuitive solution.

Rhymond:
https://github.com/laradock/laradock/pull/1293 related

PR #1294, opened 2017-12-04
fix bug on nginx container

iliasnaamane:
There is an error in the DockerFile of the Nginx container ( only for Ubuntu users, it works fine in Mac OS ). ERROR: Dockerfile parse error line 13: unknown instruction: ;FI
It should be: 
`
ARG CHANGE_SOURCE=false
RUN if [ ${CHANGE_SOURCE} = true ]; then \
    # Change application source from dl-cdn.alpinelinux.org to aliyun source
    RUN sed -i 's/dl-cdn.alpinelinux.org/mirrors.aliyun.com/' /etc/apk/repositories;fi`


Mahmoudz:
@iliasnaamane thanks for the PR, this was solved in a previous PR. So I merged #1273 instead.

PR #1327, opened 2017-12-28
Ssh for laradock user

ed-fruty:
Now ssh keys can be used only from `root` user by default.
Sometimes you will need to do something like install composer dependencies from your own private repositories via ssh keys and etc. 

So It will be good to allow it for laradock user (not root only) out of the box.

### DIFF
--- a/workspace/Dockerfile-56
+++ b/workspace/Dockerfile-56
@@ -209,6 +209,8 @@ RUN if [ ${INSTALL_WORKSPACE_SSH} = true ]; then \
         && rm -f /tmp/id_rsa* \
         && chmod 644 /root/.ssh/authorized_keys /root/.ssh/id_rsa.pub \
     && chmod 400 /root/.ssh/id_rsa \
+    && cp -rf /root/.ssh /home/laradock \
+    && chown -R laradock:laradock /home/laradock/.ssh \
 ;fi
 
 #####################################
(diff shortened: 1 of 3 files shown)


wvdongen:
I'm still unable to login using the laradock user with the specified SSH key. It does work for the root user. Any ideas?

ed-fruty:
@wvdongen can you describe what you want to do?
This PR copied ssh keys from `root` to `laradock`. So it assume you can use your private repositories from your own gitlab for instance. Or another one is access as `laradock` user from workspace container to another servers via ssh and etc. So it's not necessary to store any pub keys in your `/home/laradock/.ssh/authorized_keys` for such kind of task. You will need to store content of `/home/laradock/.ssh/id_rsa.pub` in the settings of your repository, to allow `laradock` user access for your private source.
And it's only as outbox configuration. If you need more security you can regenerate ssh keys.

wvdongen:
@ed-fruty I’m trying ssh in to the workspace with the laradock user so I can use my IDE to work directly in the workspace container. Node and composer require to use the laradock user. I’ve also tried putting my own key in authorized_keys, and I played around with the sshd config, but I can’t get it to work for the laradock user.

wvdongen:
@jasperf I know how I can su to the laradock user, that's something completely different from what I'm trying to achieve. I need to SSH as the laradock user to get my IDE working: https://github.com/laradock/laradock/pull/1327#issuecomment-364740606

mariano-dagostino:
@wvdongen This was fixed by this PR #1399. That PR also updates the [documentation](http://laradock.io/documentation/#access-workspace-via-ssh) to reflect how you can login using a the laradock user.

wvdongen:
@mariano-dagostino thanks, just tested it and it solved my problem.

This issue can be closed.

PR #1330, opened 2017-12-28
add nginx openssl

gueroverde:
allow install openSSL  when needed  to self signed certificed

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -130,11 +130,13 @@ services:
         args:
           - PHP_UPSTREAM_CONTAINER=${NGINX_PHP_UPSTREAM_CONTAINER}
           - PHP_UPSTREAM_PORT=${NGINX_PHP_UPSTREAM_PORT}
+          - NGINX_SSL_INSTALL=${NGINX_SSL_INSTALL}
       volumes_from:
         - applications
       volumes:
         - ${NGINX_HOST_LOG_PATH}:/var/log/nginx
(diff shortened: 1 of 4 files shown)


Shadow2004:
Your code is missing the part that actually generates the certificates, unless i am missing something.

What you would want in the nginx docker file is something like the below

```bash
ARG NGINX_SSL_INSTALL=false
ENV NGINX_SSL_INSTALL ${NGINX_SSL_INSTALL}
RUN if [ ${NGINX_SSL_INSTALL} = true ]; then \
    apk add --no-cache openssl  \
    && apk update \
    && mkdir /etc/nginx/ssl \
    && openssl genrsa -out "/etc/nginx/ssl/${DOMAIN_NAME}.key" 2048 \
    && openssl req -new -key "/etc/nginx/ssl/${DOMAIN_NAME}.key" -out "/etc/nginx/ssl/${DOMAIN_NAME}.csr" -subj "/CN=${DOMAIN_NAME}/O=${COMPANY_NAME}/C=${COUNTRY_CODE}" \
    && openssl x509 -req -days 365 -in "/etc/nginx/ssl/${DOMAIN_NAME}.csr" -signkey "/etc/nginx/ssl/${DOMAIN_NAME}.key" -out "/etc/nginx/ssl/${DOMAIN_NAME}.crt"
;fi
```

gueroverde:
Right just that is what I wanted to achieve @Shadow2004

PR #1345, opened 2018-01-09
workspace container adds swoole

jdxia:
workspace container no swoole extension
I added swoole extension


jdxia:
Swoole extensions are usually run in the CLI environment
so,I think the workspace container needs it

Issue #1391, opened 2018-02-23
Impossible to change MySQL version


AnkurPrem:
### Info:
- Docker version 17.12.0-ce, build c97c6d6
- Win10

### Issue:
At first I built laradock with mysql8.0 (default). After many host access problems (maybe another issue), I:

1. Changed my .env to:

`MYSQL_VERSION=5.7`

(tried to set the value in the `docker-compose.yml` directly also, same result.

2. Run:

```
docker-compose stop
docker-compose down
```

3. Deleted the `.laradock/` from the user folder (win).

3. Run:

`docker-compose build --no-cache mysql`

### Result:
```
Building mysql
Step 1/10 : ARG MYSQL_VERSION=8.0
Step 2/10 : FROM mysql:${MYSQL_VERSION}
 ---> f0f3956a9dd8
Step 3/10...
```

So, mysql 8.0 is always being built, no matter what.

Nothing I found online solves the issue.

Anyone can help me? 
Please tell me if you need more info.

AnkurPrem:
The crazy thing is that many times I didn't even use the `.env`, but the output always came `Step 2/10 : FROM mysql:${MYSQL_VERSION}`.

bestlong:
@AnkurPrem this is my mysql 5.7 build info:
```
~/Sites/laradock (master)$ docker-compose build mysql
Building mysql
Step 1/10 : ARG MYSQL_VERSION=8.0
Step 2/10 : FROM mysql:${MYSQL_VERSION}
 ---> f008d8ff927d
Step 3/10 : MAINTAINER Mahmoud Zalt <mahmoud@zalt.me>
 ---> Using cache
 ---> 0120e469e73d
Step 4/10 : ARG TZ=UTC
 ---> Using cache
 ---> 61849f836e9d
Step 5/10 : ENV TZ ${TZ}
 ---> Using cache
 ---> d68bfcd4a6f8
Step 6/10 : RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 ---> Using cache
 ---> bfb45d787e0c
Step 7/10 : RUN chown -R mysql:root /var/lib/mysql/
 ---> Using cache
 ---> e451c82da146
Step 8/10 : ADD my.cnf /etc/mysql/conf.d/my.cnf
 ---> Using cache
 ---> 647f45691866
Step 9/10 : CMD ["mysqld"]
 ---> Using cache
 ---> 266b8503ed64
[...]

PR #1409, opened 2018-03-02
Removes hard version for MySQL.

eltiare:
This allows for installing older versions of MySQL, this brings the code in alignment with current docs.

- [X] I've updated the **documentation**. (refer to [this](http://laradock.io/contributing/#update-the-documentation-site) for how to do so).

### DIFF
--- a/mysql/Dockerfile
+++ b/mysql/Dockerfile
@@ -1,4 +1,4 @@
-ARG MYSQL_VERSION=8.0
+ARG MYSQL_VERSION
 FROM mysql:${MYSQL_VERSION}
 
 MAINTAINER Mahmoud Zalt <mahmoud@zalt.me>


bestlong:
Why not use latest or 5.6/5.7 version?

eltiare:
This lets people decide which version they're going to use, as is stated 
in the docs. If you hard code it in the Dockerfile then they can't 
specify which is installed in the .env file.

eltiare:
Right now this only installs 8.0, which is a developmental version.

bestlong:
please refer #1391 

In Dockerfile ARG instruction has a default value is better then nothing.

even if change .env file `MYSQL_VERSION=8.0` to `MYSQL_VERSION=5.7`
run `docker-compose build` always show 8.0 info
real version please use `docker inspect <image id>` command to check.

eltiare:
I couldn't get it to install 5.7 until I changed this. I ran it multiple times, removing the containers before each attempt. I kept getting version 8 installed (both reported and in all actuality).

bestlong:
#1426

ryancwalsh:
I think https://github.com/laradock/laradock/issues/1392#issuecomment-383631421 fixed it for me.

PR #1427, opened 2018-03-15
workspace Dockerfile remove duplicate command in User Aliases section.

bestlong:


bestlong:
Oh, my bad. this is not duplicate, is set user alias both.

PR #1466, opened 2018-04-04
Add missing syntax to continue line

fenech:


idbartosz:
Issue #1467 - fix works.

PR #1487, opened 2018-04-18
run php-fpm under workspace user

lordofthebrain:
Run php-fpm under the workspace user. This solve all permissions problems. All files like cache files will be created with the same user

I have also add -l flag to useradd. When you use laradock with a domain user you have a large uid and your machine will be frezze by useradd without -l flag. usermod has no -l flag so I use a sed hack in php-fpm Dockerfile to avoid the frezze problem.
 
See: https://github.com/moby/moby/issues/5419 or
https://forums.docker.com/t/run-adduser-seems-to-hang-with-large-uid/27371


jasperf:
Found this an interesting idea as I thought GD Library perhaps had issues resizing images due to permissions so I tried. Longer story at https://github.com/laradock/laradock/issues/1969 . Did not see this helping changing image files to being stored with permissions `laradock:laradock` though. And error 500 remained. So perhaps this pull request is not working this way.

lanphan:
already have.
Thanks for your PR.

PR #1509, opened 2018-04-27
Added New Relic APM and fixed Travis build errors

chrisp-github:
Two things to note, one i had to move the swoole install in the workspace to the bottom after the PHP version check else i was experiencing segmentation faults in other packages installs if it happened before them. If someone has a better solution to this please give me a shout.

Second I removed the pdo install from the php-worker as this was causing a build error when building under PHP 7.0 and gets installed anyway with the pdo_mysql and pdo_pgsql, unless im wrong on that and some one give me a shout if that is the case.


Shadow2004:
So any time line on when this might get accepted as the New relic code in this is the last custom element that my fork uses and the build fixes should fix most of the build errors other people are having?

makowskid:
Massively outdated PR, we'd need a completely new version of it to reconsider.

Issue #1653, opened 2018-06-26
Run my personal docker into "workspace" docker


vlauciani:
Dear all

I've some API that run docker container.
Is there a way to "build" and "exec" my personal docker into "workspace" docker? How I can run it?

Thank you.

vlauciani:
ok, with `dind` it works perfectly!

PR #1655, opened 2018-06-27
Update docker-compose.yml to add 'dind'

vlauciani:
Add 'dind' image to run docker-in-docker: issue https://github.com/laradock/laradock/issues/1653

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -149,10 +149,13 @@ services:
         - "dockerhost:${DOCKER_HOST_IP}"
       environment:
         - PHP_IDE_CONFIG=${PHP_IDE_CONFIG}
+        - DOCKER_HOST=tcp://docker-in-docker:2375
       depends_on:
         - workspace
       networks:
         - backend
+      links:
+        - docker-in-docker
 
 ### PHP Worker ############################################
     php-worker:
@@ -714,3 +717,13 @@ services:
       ports:
         - 9010:9000
       
+### Docker-in-Docker ################################################
+    docker-in-docker:
+      image: docker:dind
+      privileged: true
+      volumes:
+        - ${APP_CODE_PATH_HOST}:${APP_CODE_PATH_CONTAINER}
+      expose:
+        - 2375
+      networks:
+        - backend


hackel:
@vlauciani What was the purpose of adding this?  I can't find any mention of it in the documentation.  I have a hard time imagining why php-fpm, for example, would need to manage its own docker containers, and the security implications frighten me.  Perhaps I'm missing something, though.  At the very least there should probably be a warning about how insecure this is.

vlauciani:
Hi @hackel

I work in a research Italian Institute and we are “deckerizing” our procedures, scripts and software.
Using Laradock I’m build a services on top of these dockers but to do that I need “dind” and the container that run various docker in Laradock is php-fpm.

For example if I would run a simple “hello world” docker from Laravel framework using Laradock you I to:
- deploy the “hello-world” docker image into “php-fpm” container
- write a simple route in Laraval that run a container from “hello-word” image
- get the output and return it to the user
- destroy the container. 

Will be my care to update the documentation.

lanphan:
Hi,
I think using dind is an optional extension, not a "must have". Is it possible to configure it using env?

Thanks.

cwilby:
While I can see that @vlauciani has a valid use case for this feature, I'm surprised this feature was merged _as a default for all Laradock users_ without questioning the security of giving php-fpm access to the Docker runtime.  

I realize that the onus is on developers to configure their services correctly, but framework developers also have a responsibility to provide as much security as they can.

I think this needs to be turned off by default, with the option to turn it on for these (assuming) rare use cases.

The security mindset needs to be so much better here.

dlnsk:
@vlauciani sorry but I can't understand how to deploy "hello-world" container into "php-fpm" and run it from. Could you explain it? Thanks.

vlauciani:
Hi @dlnsk 

First of all you need to install `docker` client into `php-fpm` container:
```
docker-compose exec -T php-fpm sh -c "curl -sS https://download.docker.com/linux/static/stable/x86_64/docker-19.03.2.tgz -o /tmp/docker.tar.gz && tar -xzf /tmp/docker.tar.gz -C /tmp/ && rm /tmp/docker.tar.gz && cp /tmp/docker/docker* /usr/local/bin && chmod +x /usr/local/bin/docker*"
```

then, you can install `hello-world` with the command:
```
docker-compose exec -T php-fpm sh -c "docker run hello-world"
```

Or, as in my case, you can build your own image:
```
docker-compose exec -T php-fpm sh -c "cd my_docker && docker build --tag my_image ."
```

this is a very good solution... but I understand that `dind` feature should not be installed as a default. I'm still thinking how to turn off by default and enable if needed.

PR #1664, opened 2018-06-29
laradocker

echonlyou:


jasperf:
Why changing to image optimizers false in the example .env? Are you saying the majority of Chinese users do not use image optimization ? Odd as it is convenient and helps your website load faster, which is better UX and SEO. And to ask end users to optimize with software on PC often does not work well.

PR #1675, opened 2018-07-05
This developer needed to turn back the clock in the PHP-FPM container, you will not believe what he did next!

4c0n:
These changes allow one to install and use [libfaketime](https://github.com/wolfcw/libfaketime) in the PHP-FPM container. And allows a user to control the date/time that is returned from the system clock.

This is very useful for debugging, especially when the UI and application have conditions on the time. For example a calendar that displays 7 days and only shows the current day and 6 following days. To debug a problem discovered in production all that needs to be done in this case is copy the entire project to a local machine and change the `PHP_FPM_FAKETIME` variable (for example 5 days ago == -5d). 

- [X] I've updated the **documentation**. (refer to [this](http://laradock.io/contributing/#update-the-documentation-site) for how to do so).


4c0n:
@bestlong Thank you!

PR #1830, opened 2018-10-09
FROM must to be the first instructions. Fixing error with docker-comp…

humop:
Hi,

I was having problems running the docker-compose.yml file, I was getting two errors in the mysql Dockerfile and the workspace Dockerfile. 

- MySQL: In this case the FROM instruction was not the first instruction as it has to be.
- Workspace: In this case the instruction ARG LARADOCK_PHP_VERSION was duplicated and avoiding the FROM instruction be the first instruction.

Hope this change can be helpful.

Thank you for all your hard work.

### DIFF
--- a/mysql/Dockerfile
+++ b/mysql/Dockerfile
@@ -1,5 +1,6 @@
-ARG MYSQL_VERSION=latest
 FROM mysql:${MYSQL_VERSION}
+
+ARG MYSQL_VERSION=latest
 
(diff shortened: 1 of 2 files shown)


bestlong:
@humop 
your docker and docker-compose version is? 

please upgrade and ref. https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

humop:
Sorry my bad. Needed to update docker-composer. Thank you for the reference too. :)

PR #1851, opened 2018-10-22
added config file for beanstalkd console with correct server settings…

simlux:
…; copy config file to source (Dockerfile)

### DIFF
--- a/beanstalkd-console/config.php
+++ b/beanstalkd-console/config.php
@@ -1,0 +1,23 @@
+<?php
+$GLOBALS['config'] = [
+    /**
+     * List of servers available for all users
+     */
+    'servers' => ['beanstalkd:11300' => 'beanstalkd:11300'],
+    /**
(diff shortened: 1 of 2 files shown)


beanstalkd-console/config.php
```diff
@@ -0,0 +1,23 @@
...
+     */
+    'auth'    => [
+        'enabled'  => false,
+        'username' => 'admin',
+        'password' => 'password',
+    ],
+    /**
+     * Version number
+     */
+    'version' => '1.7.10',
```
vladyslavstartsev: I think it should not be hardcoded that way, because currently we are pulling the latest version from `master`
`https://github.com/ptrofimov/beanstalk_console/archive/master.tar.gz`

simlux:
You're right!! I've overlooked the fixed version. I will look how I can use the current version in the config file. Thanks!

PR #1852, opened 2018-10-23
Add logstash support

### LINKED ISSUE #1321: add logstash to stack of laradock

blunket:
This PR adds support for Logstash. I have not updated documentation yet at this time as I want to ensure this is on the right track first.

This PR is a product of my first-ever experience with any of Docker, Laravel, Laradock OR Logstash, so I expect I may have made mistakes here. I would love any help and feedback to make this as perfect and portable as possible to help other devs. (Including help making anything into an environment variable that should be)

This fixes #1321 also.


docker-compose.yml
```diff
@@ -697,7 +715,7 @@ services:
       networks:
         - frontend
         - backend
-
```
vladyslavstartsev: can you please revert all of those whitespace removing lines back?
This PR is small, but it's definitely  easier to look at PR without such things

vladyslavstartsev:
@blunket there is also not that much activity in past months

blunket:
hey, thanks for the merge! I was expecting some changes would be necessary first. For example the people using logstash may not need MySQL. should we account for that or are we good for now?

bestlong:
Just take up more storage space.
If you think the removal will be more appropriate.
new PR is welcome.

DJviolin:
@blunket Please can you explain why the Logstash configuration needs to include `env_file` field explicitly?

It's creating an issue if you want to create a Laradock folder structure, where you keep it's repository untouched and calling docker-compose from a parent folder, it throws error because of file missing, for example:

```sh
$ docker-compose --env-file .env -f laradock/docker-compose.yml -f docker-compose.override.yml config up -d nginx mariadb adminer
```

Related issue: [https://github.com/laradock/laradock/issues/3146](https://github.com/laradock/laradock/issues/3146)

DJviolin:
I think it would be much better to only include the needed environment variables for Logstash, like every other images. This way, it will be possible to use Laradock without to touch any file within the Laradock git repository and only copy out the .env file to the parent dir, where calling docker-compose is happening.

PR #1854, opened 2018-10-25
laravel-horizon

frnwtr:
Added laravel horizon


xlstefan:
Hi @frnwtr ,

Can you explain a bit how do I get this container to run horizon correctly?
Status of horizon when I go to /horizon says:  Inactive...
What am I missing?

xlstefan:
Ah, just needed to rename conf file:
laravel-horizon.conf.example -> **laravel-horizon.conf**

fridzema:
Is there any reason to install memcached in horizon container? 
Horizon only works with redis.

xlstefan:
Exactly, it doesn't have sense to have memcached then.
Container should contain only the necessary stuff.

vesper8:
so.. someone should make a PR to remove memcached then ?

I'm reading elsewhere that laradock has a reputation for carelessly bloating up its containers.. I guess this is what they mean

PR #1866, opened 2018-11-04
Remove duplicated docker-sync-strategy

JamShady:
Removed the duplicate DOCKER_SYNC_STRATEGY values from env-example that was unnecessarily added in the Thumbor merge via PR #1373 

I note the existing PR #1833, but that one removes the original value which is higher up in the file in the general setup area, and thus more visible and appropriately placed for developers tweaking their config, as opposed to this duplicated entry which appears near the middle of the file in the container configuration options, and you'd have to know about it to change it.


bestlong:
duplicate #1808

PR #1888, opened 2018-11-22
FIX PHP YAML extension #1798

michaelradionov:
Just changed parameter's name in .env for PR #1798


michaelradionov:
Argh, one more mistake. Thanks for fixing!

PR #1926, opened 2019-01-02
`apt-get update` needs to be run for mssql to work properly

fredwu:
Hi,

We found that `apt-get update` needs to be run before pulling in `apt-transport-https` otherwise `pdo_sqlsrv` won't be added successfully.


bestlong:
this issue cause by build cache.
Can U build php-fpm with no cache, use `docker-compose build --no-cache php-fpm`

PR #1930, opened 2019-01-02
add new conf example file to run laravel scheduler in php-worker

lanphan:
php-worker: new example conf to start Laravel Scheduler in php-worker. This configuration is verified and run well on our system.

How it run?
+ Using Supervisord to start process forever
+ In this process, start a new background process to run schedule:run every minute. This background process will be reaped by supervisord after done.

### DIFF
--- a/DOCUMENTATION/content/documentation/index.md
+++ b/DOCUMENTATION/content/documentation/index.md
@@ -532,30 +532,37 @@ phpunit
 <a name="Run-Laravel-Queue-Worker"></a>
 ## Run Laravel Queue Worker
 
-1 - First add `php-worker` container. It will be similar as like PHP-FPM Container.
-<br>
-a) open the `docker-compose.yml` file
-<br>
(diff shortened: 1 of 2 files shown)


Surt:
Why is the scheduler runned as a supervisor process? Why not in cron? I mean, is there any preference here? must I use this instead of cron for any reason?

lanphan:
Workspace is just a place to run command composer, artisan, npm ... and you can completely stop it on production. With this in mind, I chose worker container as a place to keep all back ground jobs (cron, worker).

Surt:
> Workspace is just a place to run command composer, artisan, npm ... and you can completely stop it on production. With this in mind, I chose worker container as a place to keep all back ground jobs (cron, worker).

yes, thats right, on production what I did was moving the cronjobs with schedule to the worker crond (wich I added to the dockerfile)

but after all I prefer your solution of using supervisor since I can specify the user running the scheduler without the hassle of user permissions etc for cron on alpine.  

Thanks!

PR #2046, opened 2019-03-16
update adminer to support SQL Server

lanphan:
Update adminer to suport SQL Server:
1. Use mssql volume, don't use data volume
2. Use latest version of pdo_sqlsrv
3. Already checked and worked well with mssql container

- [ x I enjoyed my time contributing and making developer's life easier :)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -359,7 +359,7 @@ services:
         - SA_PASSWORD=${MSSQL_PASSWORD}
         - ACCEPT_EULA=Y
       volumes:
-        - ${DATA_PATH_HOST}/mssql:/var/opt/mssql
+        - mssql:/var/opt/mssql
       ports:
         - "${MSSQL_PORT}:1433"
       networks:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -359,7 +359,7 @@ services:
         - SA_PASSWORD=${MSSQL_PASSWORD}
         - ACCEPT_EULA=Y
       volumes:
-        - ${DATA_PATH_HOST}/mssql:/var/opt/mssql
+        - mssql:/var/opt/mssql
       ports:
```
bestlong: please revert this part.
all DB container best use the same way,

docker-compose.yml
```diff
@@ -359,7 +359,7 @@ services:
         - SA_PASSWORD=${MSSQL_PASSWORD}
         - ACCEPT_EULA=Y
       volumes:
-        - ${DATA_PATH_HOST}/mssql:/var/opt/mssql
```
bestlong: Please revert this part.
all DB container best use the same way,

lanphan: @bestlong using ${DATA_PATH_HOST}, it cannot run if running on Mac. Should we have some comment and example here to let them know?

PR #2123, opened 2019-05-09
Cannot load Xdebug - it was already loaded

manierim:
xdebug.so is probably already activated in `20-xdebug.ini` so there's no need to override the phpunit alias

also `/var/www/vendor/bin/phpunit` instead of `./vendor/bin/phpunit` breaks phpunit, pu, puf and pud when vendor folder is not in root app directory


vicenteherrera:
+1 to this.
phpunit as an alias for /var/www/vendor/bin/phpunit also makes impossible to make path look for it unless you unalias it.

Issue #2169, opened 2019-06-10
PHP Warning: PHP Startup: Unable to load dynamic library 'tideways.so'


vicenteherrera:
### Info:
- Docker version (`$ docker --version`): Docker version 18.09.2, build 6247962
- Laradock commit (`$ git rev-parse HEAD`): 63fc1fde44cc00aacb89511a441cb83685e8010c
- System info (Mac, PC, Linux): Mac & PC/Windows
- System info disto/version: macOS 10.14 Mojave & Windows 10

### Issue:
Each PHP execution returns a warning about missing tideways.so extension (used for XHPROF), when INSTALL_XHPROF=false in .env:
`PHP Warning:  PHP Startup: Unable to load dynamic library 'tideways.so'`

_____

### Expected behavior:
No warning message
_____

### Reproduce:
```
$ cd laradock
$ ./php-fpm/xdebug status
PHP Warning:  PHP Startup: Unable to load dynamic library 'tideways.so' (tried: /usr/local/lib/php/extensions/no-debug-non-zts-20170718/tideways.so (/usr/local/lib/php/extensions/no-debug-non-zts-20170718/tideways.so: cannot open shared object file: No such file or directory), /usr/local/lib/php/extensions/no-debug-non-zts-20170718/tideways.so.so (/usr/local/lib/php/extensions/no-debug-non-zts-20170718/tideways.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0
PHP 7.2.15 (cli) (built: Feb 21 2019 23:48:47) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.15, Copyright (c) 1999-2018, by Zend Technologies
```
_____

### Relevant Code:

```
// Shows on any php code execution
```

vicenteherrera:
For a more permanent solution, see pull request #2170 that edits laradock/php-fpm/Dockerfile.

A workaround way to fix it is to execute:
```
cd laradock
docker-compose exec php-fpm sed -i "s/extension=tideways.so/;extension=tideways.so/" /usr/local/etc/php/conf.d/xhprof.ini
```

But if you rebuild the php-fpm image, you will have to run that again.

forecho:
#2141

JexPY:
Core Warning: PHP Startup: Unable to load dynamic library 'tideways.so' Still getting error after update.

PR #2170, opened 2019-06-10
Fixed warning about tideways.so not loaded if INSTALL_XHRPOF=false in .env

vicenteherrera:
A fix to #2169 :

Each PHP execution returns a warning about missing tideways.so extension (used for XHPROF), when INSTALL_XHPROF=false in .env:
`PHP Warning: PHP Startup: Unable to load dynamic library 'tideways.so'`

You can check this just using:
```
$ cd laradock
$ ./php-fpm/xdebug status
```
(There is no documentation about INSTALL_XHRPOF setting, so no update to docs provided)

### DIFF
--- a/php-fpm/Dockerfile
+++ b/php-fpm/Dockerfile
@@ -278,9 +278,10 @@ RUN if [ ${INSTALL_XHPROF} = true ]; then \
     ) \
     && rm -r xhprof \
     && rm /tmp/xhprof.tar.gz \
+    && COPY ./xhprof.ini /usr/local/etc/php/conf.d \
 ;fi
 
-COPY ./xhprof.ini /usr/local/etc/php/conf.d
+
 
 ###########################################################################
 # AMQP:


vicenteherrera:
Checked "Updated the documentation" to represent that no documentation exists, so it can't be updated.

hackel:
Did you even bother testing this?  You're mixing Dockerfile COPY with a shell command.  Instead, I would leave the COPY line as-is, but remove the `extension=` lines from xhprof.ini.  Then add `echo extension=tideways.so > /usr/local/etc/php/conf.d/php-ext-tideways.ini` inside the if statement.

vicenteherrera:
> Did you even bother testing this? You're mixing Dockerfile COPY with a shell command. Instead, I would leave the COPY line as-is, but remove the `extension=` lines from xhprof.ini. Then add `echo extension=tideways.so > /usr/local/etc/php/conf.d/php-ext-tideways.ini` inside the if statement.

I did test it, and use it daily. My only change was moving one line of code from the outside of the IF, to the inside of it, anything else is as I found it. Any further improvements you make will be welcomed ;-)

php-fpm/Dockerfile
```diff
@@ -278,9 +278,10 @@ RUN if [ ${INSTALL_XHPROF} = true ]; then \
     ) \
     && rm -r xhprof \
     && rm /tmp/xhprof.tar.gz \
+    && COPY ./xhprof.ini /usr/local/etc/php/conf.d \
```
ahaurat: I believe @hackel  is correct and the correct solution would be to leave the `COPY` statement outside the `if` and change this line to the following:
```
&& echo extension=tideways.so > /usr/local/etc/php/conf.d/php-ext-tideways.ini \
```

You will also want to comment out (or remove) `extension=tideways.so` from `php-fpm/xhprof.ini`.

forecho:
#2141

millman2394:
Thanks for bringing up this issue. I was struggling with it using Craft CMS 3. 

I did what @ahaurat said and added `&& echo extension=tideways.so > /usr/local/etc/php/conf.d/php-ext-tideways.ini \` to the dockerfile and commented out the line in the .ini file

makowskid:
merged in different PR

PR #2179, opened 2019-06-20
Update to Elasticsearch 7.1.1

airguillaume:
Hi, 

I have updated Elasticsearch to the 7.1.1 version. I added in the docker-compose file two environnement variables that need to be there because of [breaking compatibilities](https://www.elastic.co/guide/en/elasticsearch/reference/7.0/breaking-changes-7.0.html#breaking_70_discovery_changes)

Hope this is useful for others.

Best regards,
Guillaume R.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -664,7 +664,7 @@ services:
         - node.name=laradock-node
         - bootstrap.memory_lock=true
         - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
-        - "cluster.initial_master_nodes=laradock-node"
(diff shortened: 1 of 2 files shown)


lionslair:
I need to keep the same version as AWS. They don't offer 7.x yet. 

Maybe a new env variable to configure it. You should also update kibana to match the same version

liuhe9:
kibana not work.
Kibana should be configured to run against an Elasticsearch node of the same version. This is the officially supported configuration.
or running  Kibana that is newer than the version of Elasticsearch

PR #2207, opened 2019-07-18
Add apt-get update before installing imagick

neomobil:
It's required to add apt-get update before installing imagick, otherwise the build process breaks.


shov:
I checked on master everything builds correctly, It should be cache, try `docker-compose build --no-cache php-fpm workspace` @neomobil

neomobil:
Hi!
Thanks. We have tried it a lot of times without success. That's why we
decided to execute apt-get update at the point where it breaks. Than it was
ok.

Alexandr Shevchenko <notifications@github.com> ezt írta (időpont: 2019.
júl. 19., P 10:48):

bestlong:
Hi @neomobil 
after `docker-compose build --no-cache php-fpm workspace`
you must `docker-compose down` (will remove all container)
and 'docker-compose up -d nginx` again to use new image.

PR #2247, opened 2019-08-22
Volumes are created on every docker start

StefanT123:

#### I think that this PR resolves the issue of the volumes that are created on every start, and the user can end up with many unused volumes.

This is done by adding persistent volumes for docker-in-docker and graylog. Also I've added volume for mongo configdb.


StefanT123:
Hi @lanphan, I've noticed that on every docker run, more and more volumes are created, so after 10 runs, I've ended up with more than 50 volumes. 

I've made some research of why is this happening and I noticed that docker-in-docker and mongo are creating all these volumes on every run. 

So after reading their documentation I saw that mongo needs a volume for `/data/configdb`, and dind needs volume for `/var/lib/docker`.

Also, earlier I've added graylog in the laradock software, and I noticed that the volume is wrong, so that's also fixed in this PR.

lanphan:
@StefanT123 
Thanks for your explanation, it's clear now.

PR #2301, opened 2019-09-28
Fix image magic

nzulfigarov:
Fix image magic


bestlong:
you should use `docker-compose build --no-cache php-fpm`

Issue #1745, opened 2018-08-06
php-fpm container keep the same uid with workspace


yswtrue:
### Info:
- Docker version (`$ docker --version`): Docker version 18.06.0-ce, build 0ffa825
- Laradock commit (`$ git rev-parse HEAD`): 1b865dd153818643b2bbab685dd7a2155e2ae9fc
- System info (Mac, PC, Linux): Linux
- System info disto/version: CentOs 7

### Issue:

php-fpm container don't read the workspace uid env. so there are some permission problem.

### Expected behavior:

php-fpm container read workspace uid env or there is a new env varlible to modify the uid of php-fpm

### Reproduce:

_____

### Relevant Code:

```
// place a code sample here
```

PR #2305, opened 2019-09-30
Run php-fpm as different user (#1745)

wmono:
Allows php-fpm to run as a different uid/gid than the hard-coded 1000:33. This is helpful in allowing the application to write to `storage` when the host user's uid is different from www-data's uid inside the php-fpm container. Similar to 03ff791e for php-worker.


wmono:
Similar to #2204, but allows GID to be set, and allows sharing uid/gid with an existing user/group (-o).

PR #2330, opened 2019-10-16
Install supervisord in the workspace

amorZhu:


shov:
I know that supervisord is available in php-worker container, what is the reason to have it in the workspace?

amorZhu:
Because some people are used to using workspace instead of php-worker.

shov:
Some people could be used to using supervisord everywhere, couldn't they?

amorZhu:
@shov yes. Some people can use it in php-worker or in workspace.

lanphan:
I agree to have option to be able to install Supervisord in workspace. However, would you please help to the same installation steps / configurations and example files for Supervisord like in php-worker? It's better if you update document to let user know how to enable and use Supervisord in workspace as well.

Thanks.

amorZhu:
@lanphan Yes, I try to allow users to customize the configuration in the worksapce/supervisord directory.

amorZhu:
@lanphan I update document and add config.

glebgoloborodko:
Good, but how to use it in workspace now? )

Issue #2379, opened 2019-11-20
Issue with upload_max_size in import section of PHPMyadmin


bharatgupta99:
### Info:
- Docker version (`19.03.5, build 633a0ea`): 
- Laradock commit (`c3620e95c8c6174f07c593e7fee3c7a6b2bdc1d7`): 
- System info (Mac): 
- System info disto/version: 

### Issue:
The upload_max_filesize is `1024M` in phpinfo() but is not reflected in phpmyadmin's import section.
_____

### Expected behavior:
New `upload_max_filesize` should be reflected in phpmyadmin's admin section.
_____

### Reproduce:
Update the `upload_max_filesize` in `php-fpm/laravel.ini` file and removed `php-fpm` image and rebuild it. 
_____

### Relevant Code:

laravel.ini file
```
date.timezone=UTC
display_errors=Off
log_errors=On

; Maximum amount of memory a script may consume (128MB)
; http://php.net/memory-limit
memory_limit = 256M
; Maximum allowed size for uploaded files.
; http://php.net/upload-max-filesize
upload_max_filesize = 1024M
; Sets max size of post data allowed.
; http://php.net/post-max-size
post_max_size = 2048M
max_execution_time=600
default_socket_timeout=3600
request_terminate_timeout=600

```

Issue #2344, opened 2019-10-28
phpMyAdmin import db size 


akhalid4493:
when creating new droplet with Docker 5:19.03.1~3 on 18.04 and change laradock .env mysql version from latest  to 5.7
after use :
docker-compose up -d --build
docker-compose up -d nginx phpmyadmin mysql workspace

I can't import database sql file large than 2 MB.
how can I increases the phpmyadmin importing database limit !!

<img width="984" alt="Screen Shot 2019-10-28 at 3 04 30 PM" src="https://user-images.githubusercontent.com/45354705/67677697-c01bc700-f995-11e9-9cc1-f342d6be8e38.png">

slamgundam:
I also have the same issue.

llioor:
+1

bharatgupta99:
same issue.

llioor:
> same issue.

I logged into the container of phpmyadmin and mysql and I changed to properties of max post size and upload file manually to a bigger size.

Sunvers:
+1

jfm-dnd:
Hi, I had the same issue and have found a workaround with adding `./php-ini/php.ini:/usr/local/etc/php/php.ini` to pma volumes section.
I use [docker4php](https://github.com/wodby/docker4php), so pma section in docker-compose.yml file looks like:
```
pma:
    image: phpmyadmin/phpmyadmin
    container_name: "${PROJECT_NAME}_pma"
    environment:
      PMA_HOST: $DB_HOST
      PMA_USER: root
      PMA_PASSWORD: $DB_ROOT_PASSWORD
    labels:
      - "traefik.http.routers.${PROJECT_NAME}_pma.rule=Host(`pma.${PROJECT_BASE_URL}`)"
    volumes:
      - /sessions
      - ./php-ini/php.ini:/usr/local/etc/php/php.ini
```
Then create a dir /php-ini with a php.ini file at the same level of the docker-compose.yml. Adding `post_max_size=400M` to it and restart override phpmyadmin, and it works.
I recently made a reminder from the result of my research, with a structure in example [https://github.com/jfm-dnd/docker4php-multiproject](https://github.com/jfm-dnd/docker4php-multiproject). (sorry I did it in french).

ntuanphuc:
same here

Issue #721, opened 2017-03-27
How to set php max file size?


hidace:
HI, 

I'm using Laradock with Craft CMS. I'm trying to upload a file that's 70MB, but am getting a 

`file could not be uploaded, because it exceeds the maximum upload size of 16.0 MB.`

Where can I update this in Laradock? I thought it might be in the laradock/php-fpm/laravel.ini file. Here's what mine looks like now:

```
date.timezone=UTC
display_errors=Off
log_errors=On

; Maximum amount of memory a script may consume (128MB)
; http://php.net/memory-limit
memory_limit = 128M
; Maximum allowed size for uploaded files.
; http://php.net/upload-max-filesize
upload_max_filesize = 128M
; Sets max size of post data allowed.
; http://php.net/post-max-size
post_max_size = 128M
```

After updating, I stopped all containers and re-started them. Is there a step I'm missing? Thank you!

hidace:
**SOLVED:** 

So I got this working. What I had to do is also adjust my nginx configuration (in the `laradock/nginx/nginx.conf` file) by extending the `keepalive_timeout 120` (originally it was set to 15). So my `nginx.conf` file now looks like this:
```

user www-data;
worker_processes 4;
pid /run/nginx.pid;
daemon off;

events {
  worker_connections  2048;
  multi_accept on;
  use epoll;
}

http {
  server_tokens off;
  sendfile on;
  tcp_nopush on;
  tcp_nodelay on;
  keepalive_timeout 120;
  types_hash_max_size 2048;
  client_max_body_size 90M;
  include /etc/nginx/mime.types;
  default_type application/octet-stream;
  access_log /var/log/nginx/access.log;
  error_log /var/log/nginx/error.log;
  gzip on;
  gzip_disable "msie6";
  include /etc/nginx/conf.d/*.conf;
  include /etc/nginx/sites-available/*;
  open_file_cache max=100;
  charset UTF-8;
}
```

After doing this (and updating my laravel.ini settings from the post above), I stopped all running containers from the `laradock` directory by running `docker stop $(docker ps -a -q)` then deleted my nginx and php-fpm containers by running `docker rm php-fpm` and `docker rm nginx`. 

Then, I re-built my nginx and php-fpm containers to make sure there wasn't a weird caching issue that was causing the problem by running:

`docker-compose build --no-cache php-fpm`
`docker-compose build --no-cache nginx`

Afterwards, I was able to upload my 70MB file. Hope this helps!

[...]

PR #2389, opened 2019-11-30
phpMyAdmin increase upload filesize to 2G

bestlong:
#2344 
#2379 
#721


trilogo-lordee:
Hi I found a bug in this commit, go to # 2839 and I think I found a really good solution for this commit.

PR #2396, opened 2019-12-04
Add the ability to install ffmpeg on Laravel Horizon container

ludo237:

## Introduction
Inspired by this [pull request](https://github.com/laradock/laradock/pull/2118) I've updated the Laravel Horizon Dockerfile to install `ffmpeg`.

## Problem to solve
The main issue is that if you don't use `php-worker`, but you use horizon as a queue manager, and you need to process jobs that need `ffmpeg` you won't be able to process them because ffmpeg is not available inside the laravel-horizon container

Does this pull request have breaking changes? No because the `env-example` contains the parameter set to **false** by default thus the update won't break any old setup

The documentation has not been updated since I was unable to find a section dedicated to laravel horizon.


cloudberrystory:
This fix solved my problem.
Thank you so much!)

cloudberrystory:
Also if you wanna use ffmpeg you should add non root user into laravel-horizon.
If you dont do that, all your files and paths will be created by root user.
You can do it like in php-worker.

laravel-horizon Dockerfile
image

laravel-horizon args docker-compose.yml
image

ludo237:
Good catch, I'll fix the PR asap

ludo237:
@cloudberrystory thanks for the updates, I've fixed my PR.

Also ping to @Mahmoudz for review and @bestlong

Issue #2447, opened 2019-12-27
Laradock cannot run php 7.4 & mysql 5.7


mehranhadidi:
### Info:
- Docker version 19.03.5, build 633a0ea 
- Laradock commit f64360ab74fcd3546c42ff7c74cc4d8ed61e34a3 
- System info Mac 
- System info disto/version:  OSX Catalina 10.15.2

### Issue:
I cannot run the laradock with PHP 7.4 & MySQL 5.7. I get an error during the building images. 
`ERROR: Service 'php-worker' failed to build: The command '/bin/sh -c docker-php-ext-install mysqli mbstring pdo pdo_mysql tokenizer xml pcntl'`
_____

### Expected behavior:
Should build & run containers.
_____

### Reproduce:

1. Install the latest version on laradock.
2. In the .env file set php 7.4 & mysql 5.7
3. Set the local storage to ./data so it will use the laradock/data as storage
4. turn on mysql local & server on php-fpm & workspace & worker usign .env configs
5. run `docker-compose up -d --build nginx mysql redis php-worker`
_____

If you need my extact .env file: 
https://www.dropbox.com/s/mxgumo5kzl9nwo2/laradock-env-php74-mysql57.rtf?dl=0

bestlong:
Hi @mehranhadidi 
update laradock and try again.

mehranhadidi:
@bestlong Looks like it is fixed. Thank you

PR #2468, opened 2020-01-14
php 7.4 php-worker required oniguruma

bestlong:
- #2437 
- #2447


bestlong:
https://github.com/laradock/php-fpm/pull/38

PR #2474, opened 2020-01-22
Improvements on ELK Stack images

brnathanlima:
This change is only intended to minimize errors when the developer wants to use containers based on images of the ELK Stack (Elasticsearch, logstash and kibana). As recommended on the documentation, when using these softwares togheter its mandatory to use them with the same version. I created an environment variable on env.example to centralize the input of the version number, so each Dockerfile can reference it on the build process.

Reference: https://www.elastic.co/guide/en/elastic-stack/current/installing-elastic-stack.html

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -770,7 +770,10 @@ services:
 
 ### ElasticSearch ########################################
     elasticsearch:
-      build: ./elasticsearch
+      build:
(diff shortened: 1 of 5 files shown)


zeroc0d3:
We can proceed with merge @brnathanlima.
Please add your documentation in this repository... Thanks

PR #2593, opened 2020-05-20
Adding pnpm support

makowskid:
#Description
Adding support for PNPM
more info here https://pnpm.js.org/en/installation

## Motivation and Context
pnpm uses hard links and symlinks to save one version of a module only ever once on a disk. When using npm or Yarn for example, if you have 100 projects using the same version of lodash, you will have 100 copies of lodash on disk. With pnpm, lodash will be saved in a single place on the disk and a hard link will put it into the node_modules where it should be installed.

As a result, you save gigabytes of space on your disk and you have a lot faster installations! If you'd like more details about the unique node_modules structure that pnpm creates and why it works fine with the Node.js ecosystem, read this small article: [Why should we use pnpm?](https://www.kochan.io/nodejs/why-should-we-use-pnpm.html)

## Types of Changes

- [-] Bug fix (non-breaking change which fixes an issue).
- [-] Breaking change (fix or feature that would cause existing functionality to not work as expected).


makowskid:
thanks for merge! that helps a lot

PR #2660, opened 2020-07-21
Add rdkafka to workspace

rtrzebinski-usc:
## Description
Add rdkafka PHP extension to workspace.

## Motivation and Context
It allows composer install to process dependencies that requires rdkafka.so PHP extension
Without this, even though it is installed in php-fpm container, it does not work with composer in Workspace.

## Types of Changes

- [] Bug fix (non-breaking change which fixes an issue).
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected).


workspace/Dockerfile
```diff
@@ -1265,6 +1265,20 @@ RUN if [ ${INSTALL_YAML} = true ]; then \
...
 
+###########################################################################
+# RDKAFKA:
+###########################################################################
+
+ARG INSTALL_RDKAFKA=false
+
+RUN if [ ${INSTALL_RDKAFKA} = true ]; then \
+    apt-get install -y librdkafka-dev && \
+    pecl install rdkafka && \
```
imbpp123, line 1276: You need to use rdkafka version to avoid problem with accidentally version upgrade.
Rule is simple - you need to control versions of software in container.

rtrzebinski-usc: I follow the pattern used in the code adding kafka to php-fpm container, I don't really wanna make this different in terms of version, so I'd leave it as it is, thanks for the comment though :)

rtrzebinski-usc:
To repo maintainer - 'continuous-integration/travis-ci/pr — The Travis CI build failed' seems to be failing on the code not related to my changes.

workspace/Dockerfile
rogerroelofs: Thanks for the PR.  In env-example, there is `PHP_FPM_INSTALL_RDKAFKA=false` I think you need something similar targeted to workspace.  Maybe `WORKSPACE_INSTALL_RDKAFKA=false`?  Then the workspace Dockerfile will need the same `ARG INSTALL_RDKAFKA=false` that the php-fpm Dockerfile has.

Issue #2679, opened 2020-08-12
CertBot  No such file or directory


lucassmacedo:
Problem to generate cert on localhost

docker-compose up certbot

Starting laradock_certbot_1 ... done
Attaching to laradock_certbot_1
certbot_1              | Saving debug log to /var/log/letsencrypt/letsencrypt.log
certbot_1              | Plugins selected: Authenticator webroot, Installer None
certbot_1              | Unable to register an account with ACME server
certbot_1              | cp: cannot stat '/etc/letsencrypt/archive/app.test/cert1.pem': No such file or directory
certbot_1              | cp: cannot stat '/etc/letsencrypt/archive/app.test/privkey1.pem': No such file or directory
laradock_certbot_1 exited with code 1

  ### Certbot #########################################
  certbot:
    build:
      context: ./certbot
    volumes:
      - ./data/certbot/certs/:/var/certs
      - ./certbot/letsencrypt/:/var/www/letsencrypt
    environment:
      - CN=app.test
      - EMAIL="test@app.test"
    networks:
      - frontend

bestlong:
check debug log `/var/log/letsencrypt/letsencrypt.log`

lucassmacedo:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/certbot/main.py", line 520, in _determine_account
    config, account_storage, tos_cb=_tos_cb)
  File "/usr/lib/python3/dist-packages/certbot/client.py", line 182, in register
    regr = perform_registration(acme, config, tos_cb)
  File "/usr/lib/python3/dist-packages/certbot/client.py", line 214, in perform_registration
    raise errors.Error(msg)
certbot.errors.Error: The ACME server believes "test@app.test" is an invalid email address. Please ensure it is a valid email and attempt registration again.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/letsencrypt", line 11, in <module>
    load_entry_point('certbot==0.27.0', 'console_scripts', 'certbot')()
  File "/usr/lib/python3/dist-packages/certbot/main.py", line 1364, in main
    return config.func(config, plugins)
  File "/usr/lib/python3/dist-packages/certbot/main.py", line 1238, in certonly
    le_client = _init_le_client(config, auth, installer)
  File "/usr/lib/python3/dist-packages/certbot/main.py", line 641, in _init_le_client
    acc, acme = _determine_account(config)
[...]

PR #2729, opened 2020-09-29
Update certbot run-certbot.sh to allow it work with multiple domains

bawa93:
Previously, running docker-compose up -d certbot would overwrite the previous certificate stored in /var/certs

Now: the certificates will be stored with the filename containing the domain name

## Motivation and Context
It allows to create and store SSL certificates for multiple domains

### DIFF
--- a/certbot/run-certbot.sh
+++ b/certbot/run-certbot.sh
@@ -2,5 +2,6 @@
 
 letsencrypt certonly --webroot -w /var/www/letsencrypt -d "$CN" --agree-tos --email "$EMAIL" --non-interactive --text
 
-cp /etc/letsencrypt/archive/"$CN"/cert1.pem /var/certs/cert1.pem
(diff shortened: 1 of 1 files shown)


adiramardiani:
Hi @bestlong 

Many people strugle with instalation certbot on laradock, because on documentation not write step by step about it (for example enable certbot on nginx)

Like https://github.com/laradock/laradock/issues/2679 
How can I fix issue about ?
`certbot_1 | cp: cannot stat '/etc/letsencrypt/archive/app.test/cert1.pem': No such file or directory`
`certbot_1 | cp: cannot stat '/etc/letsencrypt/archive/app.test/privkey1.pem': No such file or directory`

harunthuo:
Hi @adiramardiani 

Did you ever find a solution. Currently struggling with same issue.

brnathanlima:
Facing the same scenario. It's not very clear how to use certbot with laradock.

harunthuo:
posting this here for anyone that's facing the issue. Look at this issue i raised. Got a working solution:

#2889

Issue #2285, opened 2019-09-16
rebuild failed because nginx default.key permission


Goddard:
OSError: Can not read file in context: laradock/nginx/ssl/default.key

Everything in the ssl directory is written as root.  I changed permission for ownership and group to my user and then I was able to build.

BrothelUndercover:
same issue!

ultrono:
Yep I'm getting the same via a new install from the latest version on Ubuntu. 

Changing file SSL permissions didn't work. Only way I could get the nginx container running was to comment out the SSL related settings in my nginx site config.

VesterDe:
I'm also getting this issue, although yesterday a normal build just worked.

VesterDe:
I dirty fix was to chown it to my user, but that shouldn't be needed.

NoumanArshad83:
Hello,

I got the issue, to reproduce erro

1. Build default laradoc containers using command **docker-compose up -d mysql php-fpm nginx**
2. Stop containers using **docker-compose stop**
3. open **.env** file in laradoc and rename **COMPOSE_PROJECT_NAME** value to something else and save file
4. run command again **docker-compose up -d mysql php-fpm nginx**

it will give error
**Permission denied: '/home/laradock/nginx/ssl/default.key'** which were set as **root** and permissions **xr - -**

**Solution**
Changing permission to **chmod 644** resolved issue

PR #2780, opened 2020-11-25
Fix nginx default.key permission

vlauciani:
Solved issue: 
- https://github.com/laradock/laradock/issues/2285

with solution:
- https://github.com/laradock/laradock/issues/2285#issuecomment-616225902


vlauciani:
please, check this PR

PR #2872, opened 2021-03-13
fix(amqp extension): hard-code the version for PHP 8.0

### LINKED ISSUE #2871: Implement AMQP successfull building | PHP 8.0
**Is your feature request related to a problem? Please describe.**
It's not possible to build a `workspace` image with AMQP turned on. According to [this reference](https://pecl.php.net/package-changelog.php?package=amqp) it should be installed well through the PECL. But it does not. So, we need to specify a version.

**Describe the solution you'd like**
Hard-code a version for PHP 8.0
I'd like to avoid hard-coded values, but I feel like adding an extra environment variable for PHP 8.0 is a worse solution.
We are able to move the version to the environment variable, but in this case, it's not possible to automatic builds (w/o changes any versions).

**Describe alternatives you've considered**
Move the version of the installed extension in the environment variable.

**Additional context**
[Nothing]

PavelSavushkinMix:
## Description
The version could be hard-coded in the Dockerfile due to no ability to build workspace image w/o problems from scratch for different PHP versions.
Closes #2871

## Motivation and Context
We need the ability to build a `workspace` image w/o changes any versions.

## Types of Changes
- [ ] Bug fix (non-breaking change which fixes an issue).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).


PavelSavushkinMix:
I think in the future the hard-coded value can be removed when the `beta` becomes a release.

Issue #2892, opened 2021-03-18
XDEBUG - The PR to support xdebug 3 on php8 has introduced an error!


vlauciani:
I'm using Laradock on multiple project on the same machine but after the PR:
- https://github.com/laradock/laradock/pull/2818

the port `9003` on `php-fpm` was exposed foreach container:
- https://github.com/laradock/laradock/blob/master/docker-compose.yml#L279

this generate a _port already in use_!

Need to specify the XDEBUG port to expose into `.env` file.

vlauciani:
Solved, thanks for merge!

PR #2893, opened 2021-03-18
FIX - XDEBUG - The PR to support xdebug 3 on php8 has introduced an error!

vlauciani:
This PR fix the problem described in the issue:
- https://github.com/laradock/laradock/issues/2892


andrewwongyedpay:
Why need to expose the PHP_FPM_XDEBUG_PORT?

 It should be listened by the host / debug client.

Issue #3008, opened 2021-07-01
laravel-horizon - ERROR: Service 'laravel-horizon' failed to build : Build failed


vlauciani:
From 2 days, when I build *laravel-horizon*, I got this error:
```
$ docker-compose build --no-cache --pull laravel-horizon
Building laravel-horizon
[+] Building 7.5s (8/31)
 => [internal] load build definition from Dockerfile                                                                                                       0.0s
 => => transferring dockerfile: 10.01kB                                                                                                                    0.0s
 => [internal] load .dockerignore                                                                                                                          0.0s
 => => transferring context: 2B                                                                                                                            0.0s
 => [internal] load metadata for docker.io/library/php:7.4-alpine3.14                                                                                      1.3s
 => [auth] library/php:pull token for registry-1.docker.io                                                                                                 0.0s
 => CACHED [ 1/26] FROM docker.io/library/php:7.4-alpine3.14@sha256:97a5609d5765ef51760910dc8e8567a8618ee4ec8300dd2a1f01fa2313ea9ae4                       0.0s
 => => resolve docker.io/library/php:7.4-alpine3.14@sha256:97a5609d5765ef51760910dc8e8567a8618ee4ec8300dd2a1f01fa2313ea9ae4                                0.0s
 => [internal] load build context                                                                                                                          0.0s
 => => transferring context: 96B                                                                                                                           0.0s
 => [ 2/26] RUN if [ false = true ]; then     sed -i 's/dl-cdn.alpinelinux.org/mirrors.aliyun.com/' /etc/apk/repositories ;fi                              0.3s
 => ERROR [ 3/26] RUN apk --update add wget   curl   git   build-base   libmcrypt-dev   libxml2-dev   pcre-dev   zlib-dev   autoconf   cyrus-sasl-dev   l  5.8s
------
 > [ 3/26] RUN apk --update add wget   curl   git   build-base   libmcrypt-dev   libxml2-dev   pcre-dev   zlib-dev   autoconf   cyrus-sasl-dev   libgsasl-dev   oniguruma-dev   libressl   libressl-dev   supervisor   procps:
#7 0.272 fetch https://dl-cdn.alpinelinux.org/alpine/v3.14/main/x86_64/APKINDEX.tar.gz
#7 0.498 fetch https://dl-cdn.alpinelinux.org/alpine/v3.14/community/x86_64/APKINDEX.tar.gz
#7 0.843 (1/77) Installing m4 (1.4.18-r2)
[...]

PR #3009, opened 2021-07-01
laravel-horizon - Add Alpine version in the FROM instruction

vlauciani:
To solve issue: https://github.com/laradock/laradock/issues/3008


bestlong:
no support all PHP version.

Issue #3189, opened 2022-04-05
"COMPOSER_AUTH" does not match the expected JSON schema, this may result in errors and should be resolved


vlauciani:
Hi

I'm using last version of *laradock*:
- [5028e0c37bdb30b0dbe63434a80d784d7e9c896c](https://github.com/laradock/laradock/tree/5028e0c37bdb30b0dbe63434a80d784d7e9c896c)

---

Running `composer update` into `workspace` docker container i got:
```
root@1e8a5e8af717:/var/www# composer update
"COMPOSER_AUTH" does not match the expected JSON schema, this may result in errors and should be resolved:
 - Boolean value found, but an object is required
```

the problem is that the the `COMPOSER_AUTH` environment is reserved to `composer`:
- https://getcomposer.org/doc/articles/authentication-for-private-packages.md

but it set to `false` into `docker-compose.yml` file:
```
COMPOSER_AUTH=${WORKSPACE_COMPOSER_AUTH}
```

then:
```
root@0d018b5d88f6:/var/www#
root@0d018b5d88f6:/var/www# echo $COMPOSER_AUTH
false
root@0d018b5d88f6:/var/www#
```

To fix it, I need to *unset* the boolean value:
```
root@1e8a5e8af717:/var/www#
root@1e8a5e8af717:/var/www# export COMPOSER_AUTH=
root@1e8a5e8af717:/var/www#
root@1e8a5e8af717:/var/www#
root@1e8a5e8af717:/var/www# composer update
Loading composer repositories with package information
. . .
```

**MY PROPOSAL**: Change into `.env`, `docker-compose.yml` and `workspace/Dockerfile` variable  from `COMPOSER_AUTH` to `COMPOSER_JSON_AUTH`.

vlauciani:
PR to fix the problem, here:
- https://github.com/laradock/laradock/pull/3190

da-mask:
I had the same issue @vlauciani 's solution worked a treat, running LARADOCK_PHP_VERSION=8.1 Thanks!

Geovanek:
I am have same error

Geovanek:
I am have same error

amorZhu:
I had the same issue @vlauciani 's solution worked a treat, running LARADOCK_PHP_VERSION=8.1 Thanks again!!!

vlauciani:
This issue was fixed by PR:
- https://github.com/laradock/laradock/pull/3194

that implement:
[...]

PR #3190, opened 2022-04-05
Fix (issue #3189) "COMPOSER_AUTH" bug

vlauciani:
As described here:
- https://github.com/laradock/laradock/issues/3189

fixed the use of `COMPOSER_AUTH` environment that is a reserved by `composer`.


vlauciani:
This PR was included and merged into PR:
- https://github.com/laradock/laradock/pull/3194

close

PR #3216, opened 2022-05-17
WKHTMLtoX changes for Apple silicon

dstancer:
## Description
Procedure detects the architecture and inserts the appropriate strings in package names.
This way there is no error that a wrong package was attempted to be installed.
Also added a configurable package version so that the repo does not have to be updated when the version changes.

## Motivation and Context
WKHTMLtoX could not be installed on Apple silicon machines.

## Types of Changes
+ Bug fix (non-breaking change which fixes an issue).

+ I've read the [Contribution Guide](http://laradock.io/contributing).
+ I enjoyed my time contributing and making developer's life easier :)


pathros:
It'd be useful if you added the `WKHTMLTOPDF` to the `php-worker` container as well.

In my case, I want to attach a PDF file to an email, which is **queued**. When using the `php-worker`, I get the following error:

> /usr/local/bin/wkhtmltopdf not found

More info in this [reported issue](https://github.com/laradock/laradock/issues/3219) and even more detailed in [this stackoverflow question](https://stackoverflow.com/q/72367540).

PR #3388, opened 2023-05-16
add composer alias to turn off xdebug for composer runs

BinaryKitten:
## Description
xdebug is set to be always on within the workspace, this means things like composer run slower when you've not got a listening ide open.

this update adds the composer alias so that xdebug is turned off for duration of composer commands.

## Motivation and Context
after booting the workspace and ssh-ing into the workspace environment - ran `composer install` and due to not having an ide set to listen for connections - was presented with lots of repeated messages about could not connect to debugging client
```
Xdebug: [Step Debug] Could not connect to debugging client. Tried: host.docker.internal:9000 (through xdebug.client_host/xdebug.client_port).
```
Screenshot 2023-05-16 at 19 28 36

adding this alias and rebuilding - then running the same `composer install` 
Screenshot 2023-05-16 at 19 30 21

## Types of Changes

### DIFF
--- a/workspace/aliases.sh
+++ b/workspace/aliases.sh
@@ -60,6 +60,7 @@ alias egrep='egrep --color=auto'
 
 alias art="php artisan"
 alias artisan="php artisan"
+alias composer='XDEBUG_MODE=off \composer'
 alias cdump="composer dump-autoload -o"
 alias composer:dump="composer dump-autoload -o"
 alias db:reset="php artisan migrate:reset && php artisan migrate --seed"


Mahmoudz:
Thanks — disabling Xdebug during `composer` is a genuinely nice quality-of-life idea. Since an alias would change default `composer` behavior for everyone, we'd rather not alter it by default. Closing during cleanup; a rebased opt-in version (or a documented snippet) would be very welcome. 🙏

PR #3393, opened 2023-05-24
Update Dockerfile

iiooana:
Uploading a file via the laravel framework to the storage/app/public folder nginx returned 404 but the files uploaded fine. In particular, the nginx container user who had the same permissions as laradock was missing.

### DIFF
--- a/nginx/Dockerfile
+++ b/nginx/Dockerfile
@@ -20,9 +20,16 @@ RUN apk update \
 
 RUN apk add --no-cache curl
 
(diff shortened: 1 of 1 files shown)


Mahmoudz:
Thanks for digging into the permissions issue. Changing the nginx `www-data` uid from 82 to a PUID default would shift file ownership for existing users — exactly the kind of default change we avoid. Closing as cleanup; a rebased, opt-in version would be welcome. 🙏

PR #3394, opened 2023-05-24
Add enviroment to create user in nginx container

iiooana:
## Description
Add enviroment to create user in docker file.


Mahmoudz:
Thanks for this. The nginx image sets `www-data` to a fixed uid and has no entrypoint that reads PUID/PGID, so these variables wouldn't actually take effect as written. Closing for now; a rebased PR that applies the ids at runtime would be welcome. 🙏

PR #3429, opened 2023-08-06
add postgres backup 

AliEbrahimpour:
add postgres backup

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -664,7 +664,7 @@ services:
       links:
         - postgres
       environment:
-        - POSTGRES_HOST=malekfar-postgres-1
+        - POSTGRES_HOST={POSTGRES_HOST}
         - POSTGRES_DB=${POSTGRES_DB}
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -655,6 +655,22 @@ services:
...
+### pgbackups PostgreSQL ###########################################
+    pgbackups:
+      image: prodrigestivill/postgres-backup-local
+      restart: always
+      volumes:
+        - ../backup:/backups
+      links:
+        - postgres
+      environment:
+        - POSTGRES_HOST=malekfar-postgres-1
```
makowskid, line 667: ?

makowskid (changes requested):

docker-compose.yml
AliEbrahimpour: I updated this environment, and the link must be necessary.

makowskid (approved):

PR #3467, opened 2023-10-18
add   restart: always for all service

AliEbrahimpour:
## Description
i add restart: always for all service because while server restart all service not start automaticly

## Types of Changes

- [] Bug fix


AliEbrahimpour:
hi every one add restart fot all service while server restart then all container auto stated

AliEbrahimpour:
@appleboy 
@sodabrew 
@tjlytle 
@samdark

AliEbrahimpour:
I added this topic a long time ago, and no one has reviewed it.

AliEbrahimpour:
i also add  time zone for php-fpm

AliEbrahimpour:
i have also add time zone for php worker
@bestlong

w1lldone:
Hi @bestlong,

I don’t think this was the best approach. Now I have to manually stop all my active containers before shutting down my PC, or they’ll automatically start when I turn it back on.

AliEbrahimpour:
> Hi @bestlong,
> 
> I don’t think this was the best approach. Now I have to manually stop all my active containers before shutting down my PC, or they’ll automatically start when I turn it back on.

Dear friend, I am talking about the production and not the personal system where you have the opportunity to turn off the containers and when the server is turned off, it will not ask you.

w1lldone:
I understand that you use Laradock on your production server. However, Laradock may not be the most suitable option for production environments by itself. The Laradock repository states that it is a "Full PHP development environment for Docker." Additionally, the documentation recommends creating your own docker-compose.yml file specifically tailored for production environments (https://laradock.io/documentation/#prepare-laradock-for-production)

For this reason, I believe your current method is not the best approach. You should consider creating a custom docker-compose.yml file tailored for your production needs.

AliEbrahimpour:
I think that since you are working with it on your own system, you can delete those items so that this system is suitable for production before it is run again on your system.

PR #3594, opened 2025-01-07
feat: add Selenoid and Selenoid UI services for browser automation support

easantos-si:
- Introduced Selenoid service for managing browser sessions via Docker
- Added Selenoid UI for monitoring browser sessions
- Updated docker-compose configuration to include both services
- Included documentation with usage instructions

This contribution enables browser automation and testing workflows within the Laradock environment.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -14,7 +14,7 @@ APP_CODE_PATH_CONTAINER=/var/www
 APP_CODE_CONTAINER_FLAG=:cached
 
 # Choose storage path on your machine. For all storage systems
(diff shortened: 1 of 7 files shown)


Mahmoudz:
Thanks for the Selenoid work. As-is it declares an external `integration` network that makes `docker-compose up` fail for everyone until it's manually created, and it also changes an existing default, so we're closing it. A rebased PR that keeps Selenoid fully self-contained and opt-in (without touching shared defaults) would be welcome. 🙏

PR #3632, opened 2025-07-10
feat(workspace):  add legacy OpenSSL config support to workspace container

saeedhosseiinii:
### Summary
This PR adds Legacy OpenSSL Config support to the workspace container, bringing it in line with the php-fpm container which already has this feature. This enables compatibility with older OpenSSL configurations and legacy cryptographic algorithms when using OpenSSL 3.x.

### Why This Change is Needed

- **Consistency**: The php-fpm container already supports legacy OpenSSL configuration, but the workspace container does not
- **Compatibility**: Many legacy applications and libraries require older cryptographic algorithms that are disabled by default in OpenSSL 3.x
- **Development Environment**: Developers working in the workspace container need the same OpenSSL configuration options as the php-fpm container

### Usage

Users can enable legacy OpenSSL support by setting the environment variable:

```bash
PHP_LEGACY_OPENSSL=true
```

This will affect both the php-fpm and workspace containers, ensuring consistent OpenSSL behavior across the development environment.

### Backward Compatibility

This change is fully backward compatible:
- Default behavior remains unchanged (`LEGACY_OPENSSL=false`)
- No existing functionality is affected
- Only activates when explicitly enabled by the user

### DIFF
--- a/workspace/Dockerfile
+++ b/workspace/Dockerfile
@@ -1903,6 +1903,18 @@ RUN if [ ${INSTALL_GITHUB_CLI} = true ]; then \
     && apt install gh -y \
 ;fi
 
+###########################################################################
+# Legacy Openssl Config:
+###########################################################################
+ARG LEGACY_OPENSSL=false
+
+RUN if [ ${LEGACY_OPENSSL} = true ]; then \
+  if openssl version | grep -q "OpenSSL 3"; then \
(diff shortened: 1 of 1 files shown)


Mahmoudz:
Thanks. The opt-in legacy OpenSSL flag is reasonable, but it rewrites the global `openssl.cnf` and needs a build test plus `.env`/compose wiring before it's safe to merge. Closing during cleanup; a rebased PR with that wiring is welcome. 🙏

PR #3758, opened 2026-07-17
fix: improve source list handling for China and PHP version compatibility

xiagw:


Mahmoudz:
Thanks @xiagw. The trixie part is right, 8.1+ has no /etc/apt/sources.list so CHANGE_SOURCE does nothing there.

The `>= 80` check breaks 8.0 though, it's still bullseye, so sed can't find debian.sources and the build dies. Checked 7.4/8.0/8.1/8.4.

Mind dropping the version check and just looping over both paths with `[ -f "$f" ]`? Same result, nothing to update next Debian release.

Mahmoudz:
Fixed on master in cb73399e, going a bit wider than this PR: workspace had the same problem on noble, and arm64 never got the mirror at all since it uses ports.ubuntu.com. Patching whichever sources file exists instead of keying off the PHP version, so 8.0 (still bullseye) keeps working and the next Debian bump does not break it again.

Kept your quoting fix and the sed consolidation, they were right. Thanks for spotting it, closing this one out.

